PR #1, opened 2015-01-14
Update __init__.py

nickeubank:
Added note to python documentation about infomap not accepting graphs with isolated vertices (already in C documentation here: http://igraph.org/c/doc/igraph-Community.html ).


ntamas:
Thanks!

PR #20, opened 2015-02-11
Fix MSVC error LNK2019: unresolved external symbol strcasecmp

cgohlke:


ntamas:
Thanks! Is this the only change needed to make it compile with MSVC? If so, I'll probably release a new 0.7.1-3 version on PyPI soon.

cgohlke:
The source code builds OK with MSVC. However, I disable the `BuildConfiguration` stuff and instead set `extra_compile_args`, `include_dirs`, and `libraries` manually in setup.py to match my build environment. On win-amd-py2.x there are two "Ladder graph" related segfaults at <https://github.com/igraph/python-igraph/blob/release-0.7/igraph/test/flow.py#L154> and <https://github.com/igraph/python-igraph/blob/release-0.7/igraph/test/flow.py#L169>. Otherwise the tests pass.

PR #32, opened 2015-05-15
custom font family support for vertex & edge labels

deeenes:


ntamas:
Awesome, thanks! Can you please change the name of the attribute to `font` or `font_family` instead? (I know that all label-related attributes used to start with `label_` so far, but I can hardly imagine anything else that would be affected by the font so I think that the `label_` prefix is redundant). I would probably be happier with `font`, but `font_family` is okay.

I see that you are calling `hasattr(vertex, "label_family")` - is that necessary? If I remember correctly (haven't tested it), the `vertex` and `edge` variables in the iterator should always have that attribute if you declared them in `VisualVertexBuilder` and `VisualEdgeBuilder`.

deeenes:
Yes, you are right. I removed `hasattr()`, and changed the attribute names to `font`.

ntamas:
Awesome, thanks for your contribution!

PR #39, opened 2015-07-27
width & height for vertex shapes

deeenes:
All vertex shapes accept `width` and `height` parameters, just `VisualVertexBuilder` of `DefaultVertexDrawer` does not pass them, instead gives the value of `size` for argument `width`. This behavior is preserved, but now `width` and `height` can be used the way as it was designed seemingly.


ntamas:
The original PR cannot be merged now as some files have been moved around, but I've created an equivalent patch manually, which will be part of python-igraph 0.8. Thanks a lot for your work on this!

PR #47, opened 2015-11-04
Convergence of intersect_bezier_circle limiting the number of iterations.

albertoalcolea:
There is a bug in the intersect_bezier_circle function when two vertex are overlapped. The main loop of the function never converges.

May be a good idea add a parameter to limiting the number of iterations to reach the intersection point.

This parameter covers most cases, not only the mentioned case.


ntamas:
Great, thanks a lot!

PR #52, opened 2015-12-11
Release 0.7

luav:
python-igraph release 0.7 is ported to PyPy!

There was an issue [Error installing python-igraph using pip with pypy](https://github.com/igraph/igraph/issues/861) and this is the solution!

I compiled it on Ubuntu 14.04 and worked fine on default pypy:
```
Python 2.7.3 (2.2.1+dfsg-1ubuntu0.3, Sep 30 2015, 15:18:40)
[PyPy 2.2.1 with GCC 4.8.4]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,21 @@
+
+[![](https://travis-ci.org/igraph/python-igraph.svg?branch=master)](https://travis-ci.org/igraph/python-igraph)
+
+Python interface for the igraph library
+---------------------------------------
+
+igraph is a library for creating and manipulating graphs. 
+It is intended to be as powerful (ie. fast) as possible to enable the
+analysis of large graphs. 
+
+This repository contains the source code to the Python interface of
+igraph.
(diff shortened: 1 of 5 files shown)


luav:
Travis build is failed only because of the outdated settings in this branch for the travis build, and I would not like to spend time on it. The fix is verified in the real environment.

Details about floating-point related implementation of pypy can be checked here:
https://searchcode.com/codesearch/view/2130476/?
https://searchcode.com/codesearch/view/16935669/#l-18
https://searchcode.com/codesearch/view/16935659/#l-28

just in case of any questions related to this fix.

ntamas:
That's great, thanks a lot! I'll merge this and maybe make a few tweaks here and there, but in general it seems correct for me.

Also, I'll update `tox.ini` to test things under `pypy` as well.

luav:
You are welcome and thanks for the integration!

PR #53, opened 2015-12-11
PyPy support / porting implemented

luav:
[PyPy](http://pypy.org/) support / porting implemented, which fixes this issue:  [Error installing python-igraph using pip with pypy](https://github.com/igraph/igraph/issues/861).

Details about floating-point related implementation in pypy can be checked here:
https://searchcode.com/codesearch/view/2130476/
https://searchcode.com/codesearch/view/16935669/#l-18
https://searchcode.com/codesearch/view/16935659/#l-28
just in case of any questions related to this fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,3 +10,12 @@ analysis of large graphs.
 
 This repository contains the source code to the Python interface of
 igraph.
+
+## Install
+```
+$ sudo python setup.py install
+```
+See details in [Installing Python Modules](https://docs.python.org/2/install/).
+
+## Notes
+This version of python-igraph is ported into [pypy](http://pypy.org/), [release-0.7](/luav/python-igraph/tree/release-0.7) is verified on Linux Ubuntu 14.04 with default pypy.
--- a/README.md
+++ b/README.md
@@ -18,4 +18,4 @@ $ sudo python setup.py install
 See details in [Installing Python Modules](https://docs.python.org/2/install/).
 
 ## Notes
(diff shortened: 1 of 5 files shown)


ntamas:
Note that lots of test cases are failing right now; I'm working on fixing these.

luav:
Yes, PyPy implements not all of the Python API... But a lot of functionality works fine. I think that "not implemented" parts and other failures will be fixed either by contributors in future, or by PyPy updates.

By the way, it expected that tests on PyPy are executed slower than on CPython, because JIT compilation works slower than interpretation, but is executed dozen times faster. I.e. PyPy is much faster on the second execution or when there are a lot of pure Python code to be executed (not just calls of C libs).

ntamas:
> By the way, it is expected that tests on PyPy are executed slower than on CPython, because JIT compilation works slower than interpretation, but is executed dozen times faster.

This could be the case for pure Python code, but I haven't experienced this behaviour with igraph's tests. I was running the igraph test suite in a loop 100 times and the execution times stayed the same. This is probably because most of the overhead is not attributed to interpretation or JIT compilation but the fact that [PyPy has to emulate CPython's reference counting](http://pypy.readthedocs.org/en/latest/faq.html#do-cpython-extension-modules-work-with-pypy) in order to use a CPython module. Binding to igraph via `cffi` could potentially be faster (because then the JIT compiler could optimize the `cffi` bindings).

PR #56, opened 2016-01-05
fix keyword parsing bug in decompose()

Ariki:
An easy-to-fix bug in keyword argument parsing makes it impossible to run decompose() on directed graphs


ntamas:
Thanks!

PR #60, opened 2016-01-27
Decode SVG representation as string for IPython.

friso:
Plotting in IPython is currently broken.

The `_repr_svg` method of `Plot` checks for Python2 vs. Python3 and encodes into a UTF-8 str in the former case, but leaves the return value as `bytes` in the latter. However IPython expects the result to be a string. Decoding the bytes into a UTF-8 string fixes this and shows the plot in IPython.

### DIFF
--- a/igraph/drawing/__init__.py
+++ b/igraph/drawing/__init__.py
@@ -355,7 +355,7 @@ class Plot(object):
         if hasattr(result, "encode"):
             return result.encode("utf-8")     # for Python 2.x
         else:
-            return result                     # for Python 3.x
+            return result.decode("utf-8")     # for Python 3.x
 
     @property
     def bounding_box(self):


ntamas:
Thanks a lot!

friso:
You're welcome.

For Mac users, it may make sense to back port this into a 0.7 branch as well, because there's currently no homebrew-installable igraph against which it's possible to compile python-igraph's master. I actually had to make this change to the 0.7.2-pre or 0.7.1 (forgot which one) and install that locally using pip against homebrew installed igraph-0.7.1_1 to get my env to work.

ntamas:
I'm planning to make a proper 0.7.1.post7 release on PyPI soon - once I manage to sift through the commits since the last release in `master` and decide which ones are to be merged into the `release-0.7` branch. Stay tuned :)

friso:
Great. Thanks!

Issue #75, opened 2016-05-08
publish manylinux package to pypi


abe-winter:
Love igraph, hate installing a bunch of dependencies.

Is a manylinux wheel in the roadmap?

Happy to lend a hand on this -- project maintainer, assign this to me if that makes sense.

ntamas:
To be honest, this is the first time that I've heard about manylinux wheels, but I'm glad that there's finally a proposal that should in theory allow us to compile and distribute binary packages on Linux.

I would be happy to make the Python interface of igraph manylinux-compatible, but since I don't use Linux myself any more, I definitely need a hand here. I think the easiest is if you fork the repo, add any necessary patches to make it work, and then send me a pull request.

FWIW, there are sample Dockerfiles in the `docker/` subdirectory, so you can probably use `docker/minimal/Dockerfile` as a starting point for building igraph in the Docker container provided by the manylinux project.

Let me know if you need any help with this.

abe-winter:
Awesome. Thanks for the pointer to the dockerfiles. I'll make time for this sometime this week.

abe-winter:
@ntamas this is ready for review. There are some notes in the PR on testing.

To upload wheels to pypi, the manylinux docs recommend using [twine](https://pypi.python.org/pypi/twine). If there's a CI task that uploads wheels automatically for this project, let me know and I'll edit.

janschulz:
Acording to my knowledge, twine isn't easily used in CI services (e.g. https://github.com/pypa/twine/pull/165). That said, here is a script which includes such a step: https://github.com/janschulz/package-builder/blob/master/recipes/pypandoc/build.sh

You need to add $PYPI_USER and $PYPI_PASS secure variables to the CI environment...

abe-winter:
@janschulz thanks for the pointer to your conda scripts. I assume the igraph team is on a slow release cycle and uploads to PYPI manually; I'll wait to hear from them before writing any CI.

abe-winter:
@ntamas
- to generate the wheels, run `make build-wheel && make copy-wheel` in the repo root on a machine with docker installed. It will take ~ 30 minutes; the wheels will be in `docker/wheelhouse`.
- `make test-wheel` will run your test suite in various linuxes using the generated wheels

ntamas:
Tried this and it seems to work great, thanks! I'll merge the pull request soon.

ntamas:
[...]

PR #76, opened 2016-05-13
Manylinux

abe-winter:


abe-winter:
This creates manylinux wheels for igraph as per #75.

Issues:

* libxml2 should be linked statically according to the (manylinux official dependencies)[https://www.python.org/dev/peps/pep-0513/#the-manylinux1-policy] list. That said, dynamic `.so` deps work on centos & ubuntu so this isn't high priority
* libm has to be linked statically to support alpine linux (the go-to distro for small docker images)
* my 'test' is to import igraph; if there's a more comprehensive test, edit the `test-*` targets in Makefile

ntamas:
`python -m igraph.test.__init__` should run the whole igraph test suite; this is a more comprehensive test than simply trying to import `igraph`.

PR #119, opened 2017-03-17
Add commas after "i.e." in description lines.

xuhdev:
In this way, epydoc will not truncate words after "i.e." from the function summary line.


ntamas:
Thanks! Looks like `epydoc` has a strong preference towards American style :)

PR #148, opened 2017-11-14
Isolate graph drawing result to avoid id collision

### LINKED ISSUE #145: Graph visualizations output can not be used on a same page
Steps to reproduce:
1. Install latest release of igraph (python-igraph-0.7.1.post6) on python2 and open a jupyter notebook (pip install jupyter), open `jupyter notebook`, create a notebook.
2. Add these line on separate cells:
```python
import igraph as ig
ig.plot(ig.Graph.Full(3), vertex_label=['a', 'b', 'c'], bbox=(100, 100))
ig.plot(ig.Graph.Full(6), vertex_label=range(1, 7), bbox=(100, 100))
```
screen shot 2017-11-09 at 2 26 36 pm

Actual:
As can be seen on the screenshot, second graph on the page is messed up. As the both are SVGs they had some interference it seems.

Expected:
Each should drawn separately and correctly.

ebraminio:
fixes #145

This change which is suggested by @takluyver forces jupyter to isolate
graphs inside a notebook so we won't have collision of similar ids
can be available on cairo renderer.

### DIFF
--- a/igraph/drawing/__init__.py
+++ b/igraph/drawing/__init__.py
@@ -356,9 +356,11 @@ class Plot(object):
         # Return the raw SVG representation
         result = io.getvalue()
         if hasattr(result, "encode"):
-            return result.encode("utf-8")     # for Python 2.x
+            result = result.encode("utf-8")   # for Python 2.x
         else:
-            return result.decode("utf-8")     # for Python 3.x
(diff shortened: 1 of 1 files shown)


ntamas:
Seems good, thanks!

PR #151, opened 2017-11-15
Improve _is_running_in_ipython logic, fixes #149

### LINKED ISSUE #149: igraph doesn't understand ipywidgets interactive mode of jupyter
Put this on a jupyter notebook,

```python
import igraph as ig
import cairo
from ipywidgets import interact

@interact(n=(0, 10))
def interactive(n):
    return ig.plot(ig.Graph.Full(n))
```

Scroll the widget, every new graph will be opened on a separate system window at least on macOS.

I should turn above to this to avoid opening of images on separate window,

```python
import igraph as ig
import cairo
from ipywidgets import interact

mockSurface = cairo.ImageSurface(cairo.FORMAT_ARGB32, 64, 64)

@interact(n=(0, 10))
def interactive(n):
    return ig.plot(ig.Graph.Full(n), target=mockSurface)
```

ebraminio:
For some reasons it seems the current code is not able to detect ipython
when ran from ipywidgets interactive widgets, lets change it to
what @takluyver from jupyter team is suggested below:

https://github.com/jupyter/jupyter/issues/299#issuecomment-344540261

### DIFF
--- a/igraph/utils.py
+++ b/igraph/utils.py
@@ -39,11 +39,9 @@ def _is_running_in_ipython():
     """Internal function that determines whether igraph is running inside
     IPython or not."""
     try:
-        # get_ipython is injected into the Python builtins by IPython so
-        # this should succeed in IPython but throw a NameError otherwise
-        get_ipython
-        return True
-    except NameError:
+        from IPython import get_ipython
+        return get_ipython() is not None
+    except ImportError:
         return False
 
 


ebraminio:
For some reasons my local jupyter ipywidgets is broken on working interactively but this seem works on presence of ipython just like the previous approach.

ntamas:
Seems good, thanks!

PR #161, opened 2018-02-10
make print statements python3 compatible

yy:
Adding the parentheses in print statements to make code examples compatible with python3.


ntamas:
Thanks!

PR #181, opened 2018-06-13
ConfigParser instead of deprecated SafeConfigParser; +2 Linux image viewers

deeenes:


ntamas:
Thanks for this. As far as I know, `SafeConfigParser` is not deprecated yet in the 2.x series, and it was actually `SafeConfigParser` that was renamed to `ConfigParser` in Python 3.2. So, we should keep on using `SafeConfigParser` in Python 2.x and use `ConfigParser` in Python 3.x to avoid the deprecation warning. Can you do something like this?

```python
import sys
if sys.version_info < (3, ):
    from ConfigParser import SafeConfigParser as ConfigParser
else:
    from configparser import ConfigParser
```

(Also note the capitalization of the module name in Python 2.x vs Python 3.x)

ntamas:
Thanks!

PR #187, opened 2018-09-15
remove redundant definition of TextAlignment

ReblochonMasque:
the `class TextAlignment` was repeated in the code


ntamas:
Nice catch, thanks!

PR #189, opened 2018-09-24
add linux installation instructions

gchilczuk:
I've spent way too much time on searching why _Could not download and compile the C core of igraph_ so I want to help others to avoid wasting time and add a few helpful instructions. I've written down a few commands that probably solved my problems - but I'm not sure which of them are really necessary.

I don't say that my proposition is the best way, I just want the documentation to become a little better for new users. I'm looking forward to review and advice how can I make my PR better :)


ntamas:
Thanks a lot!

PR #194, opened 2018-10-08
Add support for coloring to BLISS functions

frederik-h:
Although the functions for computing canonical permuations
and testing isomorphism using BLISS in igraph's C core
support colorings, this support is not exposed in the
python library.

This commit adds support for colorings to the
corresponding functions of the python library.

*Beware:* The tests should succeed after PR #1117 has been merged into the igraph  C core


ntamas:
Seems good, thank you!

PR #203, opened 2018-12-08
Avoid DeprecationWarning when importing SafeConfigParser

### LINKED ISSUE #202: Import of SafeConfigParser below python3.2 causes DeprecationWarning
The following code

https://github.com/igraph/python-igraph/blob/3fec1d71f6d3160dc5416eec9e5ef677e6f9cd00/igraph/configuration.py#L32-L35

is causing the following (harmless) `DeprecationWarning` during normal usage

```
  /home/travis/build/cthoyt/nrl/.tox/py/lib/python3.6/site-packages/igraph/configuration.py:288: DeprecationWarning: The SafeConfigParser class has been renamed to ConfigParser in Python 3.2. This alias will be removed in future versions. Use ConfigParser directly instead.
    self._config = SafeConfigParser()
```

I would suggest appending https://github.com/igraph/python-igraph/blob/3fec1d71f6d3160dc5416eec9e5ef677e6f9cd00/igraph/configuration.py#L32 to check for less than python32 like 

```python
if sys.version_info < (3, 2):
    ...
```

to avoid throwing this warning during usage with modern python3

cthoyt:
Closes #202

### DIFF
--- a/igraph/configuration.py
+++ b/igraph/configuration.py
@@ -29,7 +29,7 @@ Foundation, Inc.,  51 Franklin Street, Fifth Floor, Boston, MA
 """
 
 import sys
-if sys.version_info < (3, ):
+if sys.version_info < (3, 2):
     from ConfigParser import SafeConfigParser as ConfigParser
 else:
     from configparser import ConfigParser


ntamas:
Thanks a lot!

PR #205, opened 2019-01-07
add solution from stackoverflow with link

niccokunzmann:
Problem: I could not find the instruction online.
Solution: This Stackoverflow answer worked.


niccokunzmann:
I think, we do not disagree of the value on stackoverflow, see the
upvotes. If you like to have it worded differently, please state your
conditions so I can make adjustments.

> From release 0.5, the C core of the igraph library is *not* included
in the Python distribution - you must compile and install the C core
separately.
 - https://pypi.org/project/python-igraph/

I did not know that it still downloads the core.

PR #220, opened 2019-05-13
tutorial fix: the graph is empty at the beginning

MartinoMensio:
Hi! Just a small suggestion on the tutorial.
In the current doc there is written that the graph has a single vertex, but in reality there are none.

Martino


ntamas:
Thanks a lot! The homepage will be updated with the next version, which will hopefully come soon(ish).

PR #254, opened 2019-10-16
Alternative installation/compilation method

vtraag:
This is an alternative installation method that I have in mind. The idea is that all necessary C code will be immediately included in the Python extension, instead of compiled separately as a library and then linked against. The `setup.py` is still a bit rough, and should be carefully checked, but it illustrates the idea.

The new file `bootstrap.sh` pulls all necessary sources from `vendor/source/igraph` and puts them in `src/core`. It relies on a new addition to `src/Makefile.am` in the `igraph` C core:
```
echodist: 
	$(info $(DISTFILES))
```
This allows `bootstrap.sh` to parse all necessary files and copy them over to `src/core`. Since I cannot refer to a specific commit in `igraph` for the submodule in `vendor/source/igraph` that includes this change (yet), the compilation on the CI systems will inevitably fail. The commit from `igraph` that I use for the submodule is https://github.com/igraph/igraph/commit/77daf8c3018f7d4302d3eecd38e92d6bd521dbca, with the added change to `src/Makefile.am` as indicated.

The new `setup.py` will by default simply compile all source files. The great benefit of doing it this way is that we can distribute a source package using `python setup.py build sdist`, which can then be compiled on all platforms relatively straightforward. It still requires to link to `xml2`, `z`, `m` and `gmp` (note that `gmp` should be replaced by `MPIR` on Windows). On most *nix systems, this should be quite straightforward. On Windows this is more challenging, but we should ensure precompiled binaries are available for Windows anyway. I think that installation and compilation would be simpler for most users in this case.

By including all necessary source code for the Python package in the source package, there will also be less ambiguity about compiling against the "correct" version of the `igraph` C core. The Python package no longer depends on a specific released version of the `igraph` C core, thus decoupling the release of the C core and the Python package. I think we can then be more flexible with both packages and make more regular releases.

It would still be possible to compile against an external `igraph` C core library by using `--with-external-igraph`. If specific compilation things are required for some systems or for some special cases, this would then still be possible. It would also still allow to use external libraries for the `igraph` C core (e.g. the external `glpk`). The vast majority of the users will probably not need this, but it would allow expert users to still fully configure the package.

Note that the current `setup.py` still contains some dirty things to get it to work on my machine. This should of course be corrected. Some other things may better be done differently, any comment is welcome!

I just wanted to open this PR so that we can discuss more easily about the possible benefits/downsides of this alternative approach.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,4 @@ igraph/*.so
 .tox
 docker/wheelhouse
 Pipfile
-
+src/core
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,0 +1,35 @@
+cd vendor/source/igraph/
+
+#./bootstrap.sh
+#./configure --disable-tls --disable-gmp
+#make
+
+cd src/
+
+SOURCES=`make echodist | head -1`
+SOURCES="$SOURCES ../config.h f2c/arith.h"
+headerfile=""
+for file in $SOURCES ; do
+    if [[ $file == *.y ]]; then
+        file="${file%.y}.c"
+        headerfile="${file%.c}.h"
+    elif [[ $file == *.l ]]; then
+        file="${file%.l}.c"
+    fi
+    if [[ $file == *.c ]] || [[ $file == *.cc ]] || [[ $file == *.cpp ]] || [[ $file == *.h ]] || [[ $file = *.hh ]] || [[ $file == *.pmt ]]
+    then
+        dir=$(dirname "$file")
+        mkdir -p "../../../../src/core/src/$dir"
+        cp "$file" "../../../../src/core/src/$file"
+        echo "Copy $file"
+        if [ -n "$headerfile" ]; then
+            cp "$headerfile" "../../../../src/core/src/$headerfile"
+            echo "Copy $headerfile"
+        fi
+    else
+        echo "Ignoring $file"
+    fi
+    headerfile=""
+done
+
+cd ../../../../
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,8 +1,8 @@
 cd vendor/source/igraph/
 
-#./bootstrap.sh
-#./configure --disable-tls --disable-gmp
-#make
+./bootstrap.sh
+./configure
+make
 
 cd src/
 
--- a/setup.py
+++ b/setup.py
@@ -231,86 +231,17 @@ LIBIGRAPH_FALLBACK_LIBRARY_DIRS = []
 
 ###########################################################################
 
-
-def cleanup_tmpdir(dirname):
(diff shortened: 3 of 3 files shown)


ntamas:
Sounds good - and sorry for the late reply, it's been a rough few weeks.

Question: would it be possible to change things such that this bootstrap script gets executed automatically when running `python setup.py sdist` (but is otherwise hidden in some subdirectory)? My idea would be as follows:

* Users who check out the source code of `python-igraph` from Github should not be required to run this script; it should be possible for them to simply run `git submodule update --init && python setup.py build` to build the extension.

* At the same time, users running `pip install python-igraph` should at least get a tarball that contains all the source code necessary to build the extension; in other words, the version that we publish to PyPI should contain igraph's source code as well, which could be achieved by your modifications. The easiest, no-friction way would be to somehow ensure that your script gets executed before the `sdist` step that builds the source tarball that eventually gets uploaded to PyPI when running `python setup.py sdist upload` (the typical way to publish a new version on PyPI).

iosonofabio:
Sorry for the sparse activity, I just started a new job and it's overwhelming.

In these cases I usually only push to pypi from a CI server e.g. Travis, appveyor. That lets you customize things while being very consistent. All I do when I want to release is adding a git tag and the CI does the rest. 

Otherwise we can modify the internals of sdist via subclassing and that will affect all sdist calls including ours but also other users'.

ntamas:
That's a good point - maybe we should simply attach the script to the Travis build somehow and use Travis to upload stuff to PyPI?

iosonofabio:
That's what I would do, let's hear from Vincent

vtraag:
Ensure that `bootstrap.sh` is run in one way or the other when runnning `python setup.py build` (or any of its alternatives) sounds like a good plan indeed! Subclassing to modify `sdist` sounds a bit too hacky for my taste. We could simply call `bootstrap.sh` from within the `python setup.py` files whenever the core source code is missing from the `src/core/` directory. This keeps the actual `setup.py` relatively lightweight. We should also make clear that this is only supported on *nix machines, not on Windows.

@iosonofabio, it does sounds like a good plan to automate releases to PyPi from the CI. The source distribution of course only needs to be made once, and this can be done from a linux CI. For other platforms, we would preferably like to provide pre-build wheels for users. One possibility would be to use https://github.com/joerick/cibuildwheel. Do you have any experience with that?

Using `git` tag for versioning would be best indeed. I believe there is now a sort of duplicate system, where `VERSION` in `setup.py` needs to be set manually. Could we perhaps simply switch to `versioneer`, or perhaps there are other, better alternatives?

The last two changes are (to some extent) independent from the current PR. Shall we open up two new issues for that?

ntamas:
I think this is now basically solved. Closing the PR.

Issue #72, opened 2016-03-23
get_adjacency(): add support to return scipy.sparse matrix...


eltermann:
... or another sparse format

ntamas:
Added this to the wishlist as it would indeed be a nice addition, but it's probably not high priority because it can already be done with a simple utility function like this:

```
from scipy.sparse import csr_matrix

def to_sparse(graph, weight_attr=None):
    edges = graph.get_edgelist()
    if weight_attr is None:
        weights = [1] * len(edges)
    else:
        weights = graph.es[weight_attr]
    if not graph.is_directed():
        edges.extend([(v, u) for u, v in edges])
        weights.extend(weights)
    return csr_matrix((weights, zip(*edges)))
```

iosonofabio:
This one was done in #274, closing.

PR #257, opened 2019-10-26
Add adjacency sparse export as separate function

iggisv9t:
As was requested in #72 
I think it's better to make it a separate function, because otherwise all added code will appear in `if sparse:` block. Moreover `get_adjacency` is already too long.

I have used @ntamas example code as base, but have added some speedup: 
Original code gives `1.94 s ± 4.45 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`
My speedup: `1.01 s ± 4.33 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`
Tested on iMDB recommendations graph with ~173K nodes and ~1568K edges.
There was also a little bug in example: sparse matrix constructor is not aware about matrix shape, so it must be provided explicitly.

### DIFF
--- a/igraph/__init__.py
+++ b/igraph/__init__.py
@@ -57,6 +57,8 @@ from collections import defaultdict
 from itertools import izip
 from shutil import copyfileobj
 from warnings import warn
+import numpy as np
+from scipy.sparse import csr_matrix
 
 def deprecated(message):
     """Prints a warning message related to the deprecation of some igraph
@@ -528,6 +530,34 @@ class Graph(GraphBase):
         """
         return DyadCensus(GraphBase.dyad_census(self, *args, **kwds))
 
+    def get_adjacency_sparse(self, attribute=None):
+        """Returns the adjacency matrix of a graph as scipy csr matrix.
+        @param attribute: if C{None}, returns the ordinary adjacency
+          matrix. When the name of a valid edge attribute is given
+          here, the matrix returned will contain the default value
+          at the places where there is no edge or the value of the
+          given attribute where there is an edge.
+        @return: the adjacency matrix as a L{scipy.sparse.csr_matrix}."""
+
+        if attribute is None:
+            return Matrix(GraphBase.get_adjacency(self, type))
+
+        if attribute not in self.es.attribute_names():
+            raise ValueError("Attribute does not exist")
+
+        edges = self.get_edgelist()
+        if attribute is None:
+            weights = np.ones(len(edges))
+        else:
+            weights = self.es[attribute]
+
+        N = self.vcount()
+        sparse_matrix = csr_matrix((weights, zip(*edges)), shape=(N, N))
+
+        if not self.is_directed():
+            sparse_matrix = sparse_matrix + sparse_matrix.T
+        return sparse_matrix
+
     def get_adjacency(self, type=GET_ADJACENCY_BOTH, attribute=None, \
             default=0, eids=False):
         """Returns the adjacency matrix of a graph.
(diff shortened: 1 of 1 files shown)


iggisv9t:
Ok, this fails because I'm trying to import scipy which is not included in requirements.

ntamas:
The `scipy` import should be moved inside the body of `get_adjacency_sparse`; otherwise it would not be possible to use `igraph` without having `scipy` installed.

Some unit tests should be added for `get_adjacency_sparse` that are skipped when `scipy` is not installed.

Also, I was wondering how loop edges should be handled in case of undirected graphs; right now the function will effectively double the weight of each loop edge in an undirected graph, and I'm not sure whether that's the right approach or not. (Either way, it should be compatible with how `get_adjacency()` behaves right now; in other words, `get_adjacency()` and `get_adjacency_sparse()` should return identical matrices for identical inputs).

szhorvat:
> Also, I was wondering how loop edges should be handled in case of undirected graphs;

Related:

 - https://github.com/igraph/igraph/issues/1193
 - https://github.com/igraph/igraph/issues/1103

> right now the function will effectively double the weight of each loop edge in an undirected graph, and I'm not sure whether that's the right approach or not. 

Depends on the application, but that is not what most systems do and that is not what people familiar with other graph tools would expect. The practical default is to not double the diagonal.

> in other words, `get_adjacency()` and `get_adjacency_sparse()` should return identical matrices for identical inputs

Absolutely!

szhorvat:
The C core already has sparse matrices, which use a CSR/CSC representation internally, and therefore are in principle easy to convert to a scipy sparse matrix.

https://igraph.org/c/doc/ch07.html#igraph-SparseMatrices

Would it make sense to make use of these? This would ensure consistency with other parts of igraph. It guarantees that the behaviour is the same as in the C core.

iggisv9t:
Thank you for comments!
> The scipy import should be moved inside the body of get_adjacency_sparse

That's right, I'll do that. Is it worth it to also include numpy? `np.zeros(n)` is faster then `[0] * n`
> Some unit tests should be added 

I need a time to get at testing in igraph
> Either way, it should be compatible with how get_adjacency()

Of course
> I'm not sure whether that's the right approach or not

I think it should be explicitly documented. User should know what to expect. It's not a problem to multiply or divide diagonal by 2 if you know you need to do so. I followed example code where was a duplication.
> that is not what people familiar with other graph tools would expect

Agree
> The C core already has sparse matrices

Only if someone who better then me in C wants to make it

vtraag:
I believe @ntamas mentioned that [`igraph_sparsemat_t`](https://igraph.org/c/doc/ch07.html#igraph-SparseMatrices2) is to be preferred over `igraph_spmatrix_t`, which might be phased out in the long run. The former is based on [`CSXSparse`](https://github.com/DrTimothyAldenDavis/SuiteSparse), a third-party implementation of sparse matrices.

The C function [`igraph_get_sparsemat`](https://igraph.org/c/doc/ch07.html#igraph_get_sparsemat) only allows to get unweighted sparse matrices. The current PR allows to specify an attribute to use as weights. Perhaps the same pattern as used for `get_adjacency` (see [`igraph/__init__.py`](https://github.com/igraph/python-igraph/blob/master/igraph/__init__.py#L531-L594) and [`src/graphobject.c`](https://github.com/igraph/python-igraph/blob/master/src/graphobject.c#L7342-L7375)) can be used to allow for attributes to be specified. In the long run, it might be better to directly support a weight vector in `igraph_get_adjacency` and `igraph_get_sparsemat`, but that is probably best left for the future.

ntamas:
I'm in favour or merging this into python-igraph 0.8 if the `scipy` import is moved inside the function so we don't need to import `scipy` when `igraph` is imported. We would probably also need to catch `ImportError`s around the `scipy` import and provide a nicer, user-friendly message that tells the user to install `scipy` if he/she wants sparse matrices.

Unfortunately some files have been moved around in the repo so this PR cannot be merged without rewriting it first.

@iggisv9t Do you have time to do the changes in the next week or so? If you don't, no problem, just let me know and I'll take care of it before the release of python-igraph 0.8.

iggisv9t:
@ntamas Yes, I'll do it. I think I'll find a time on this weekend.

iggisv9t:
Ok, now it looks like I just finished with my code, but I also created a little mess in git. So I need a bit of advice. I fixed a typo in branch name locally and then I occasionally pushed it as a separate branch on github. So easy way as I see it is to create a new PR for it. What do you think?

ntamas:
Yes, I think that's the easiest. I'll close this and then feel free to send a new PR.

iggisv9t:
Recreated as PR #274

PR #264, opened 2019-12-20
Reindex clusterings that were used for testing

vtraag:
This addresses the issue that the build of `python-igraph` now [fails](https://travis-ci.org/igraph/python-igraph/builds/626633793). This is because a cluster index is supplied that is outside the range of 0, ..., n - 1, where n is the number of nodes. This is now corrected.

The range checking in the C core is performed explicitly in PR [1273](https://github.com/igraph/igraph/pull/1273), which improves the error reporting.

It seems there is still an error in `igraph.test.walks.RandomWalkTests` when running the Python tests. I have not yet had time to investigate that.


ntamas:
Thanks a lot!

(The random walk tests seem to pass in the CI env).

vtraag:
Yes, not sure what was happening with the random walk. Perhaps I'll take a closer look when I find some time.

PR #273, opened 2020-01-31
Implementation of Leiden algorithm

vtraag:
This is the initial public interface of the Leiden algorithm in Python. I still need to add some test.

### DIFF
--- a/src/_igraph/graphobject.c
+++ b/src/_igraph/graphobject.c
@@ -11745,6 +11745,93 @@ PyObject *igraphmodule_Graph_community_walktrap(igraphmodule_GraphObject * self,
   return res;
 }
 
+/**
+ * Leiden community detection method of Traag, Waltman & van Eck
+ */
+PyObject *igraphmodule_Graph_community_leiden(igraphmodule_GraphObject *self,
+        PyObject *args, PyObject *kwds) {
+  static char *kwlist[] = {"edge_weights", "node_weights", "resolution_parameter",
+                           "beta", "initial_membership", "n_iterations"};
+
+  PyObject *edge_weights_o = Py_None;
+  PyObject *node_weights_o = Py_None;
+  PyObject *initial_membership_o = Py_None;
+  PyObject *res;
+
+  int ret = 0;
+  long int n_iterations = 2;
+  double resolution_parameter = 1.0;
+  double beta = 0.01;
+  igraph_vector_t *edge_weights = NULL, *node_weights = NULL, *membership;
+  igraph_bool_t start = 1;
+  igraph_integer_t nb_clusters = 0;
+  igraph_real_t quality = 0.0;
+
+  if (!PyArg_ParseTupleAndKeywords(args, kwds, "|OOddOl", kwlist,
+        &edge_weights_o, &node_weights_o, &resolution_parameter, &beta, &initial_membership_o, &n_iterations))
+    return NULL;
+
+  // Get edge weights
+  if (igraphmodule_attrib_to_vector_t(edge_weights_o, self, &edge_weights,
+    ATTRIBUTE_TYPE_EDGE)) {
+    igraphmodule_handle_igraph_error();
+    return NULL;
+  }
+
+  // Get node weights
+  if (igraphmodule_attrib_to_vector_t(node_weights_o, self, &node_weights,
+    ATTRIBUTE_TYPE_VERTEX)) {
+    igraphmodule_handle_igraph_error();
+    return NULL;
+  }
+
+  // Get initial membership
+  if (igraphmodule_attrib_to_vector_t(initial_membership_o, self, &membership,
+    ATTRIBUTE_TYPE_VERTEX)) {
+    igraphmodule_handle_igraph_error();
+    return NULL;
+  }
+
+  if (!membership) {
+    start = 0;
+    membership = (igraph_vector_t*)calloc(1, sizeof(igraph_vector_t));
+    if (membership==0) {
+      PyErr_NoMemory();
+      return 1;
+    }
+    igraph_vector_init(membership, 0);
+  }
+
+  /*********************/
+
+  ret = igraph_community_leiden(&self->g,
+                              edge_weights, node_weights,
+                              resolution_parameter, beta,
+                              start, membership,
+                              &nb_clusters, &quality);
+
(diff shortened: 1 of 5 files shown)


src/_igraph/graphobject.c
```diff
@@ -11745,6 +11745,135 @@ PyObject *igraphmodule_Graph_community_walktrap(igraphmodule_GraphObject * self,
...
+  PyObject *edge_weights_o = Py_None;
+  PyObject *node_weights_o = Py_None;
+  PyObject *initial_membership_o = Py_None;
+  PyObject *res;
+
+  int error = 0, i;
+  long int n_iterations = 2;
+  double resolution_parameter = 1.0;
+  double beta = 0.01;
+  igraph_vector_t *edge_weights = NULL, *node_weights = NULL, *membership;
```
ntamas: I think we could simply allocate an `igraph_vector_t membership` vector statically, no need for a dynamic allocation here. This would simplify the body of the function a bit.

ntamas: Scratch that, I was wrong, sorry. The initial membership vector comes from the arguments as a Python object so we need this to be a pointer.

src/_igraph/graphobject.c
```diff
@@ -11882,7 +12011,7 @@ struct PyMethodDef igraphmodule_Graph_methods[] = {
    "Counts the number of edges.\n"
    "@return: the number of edges in the graph.\n" "@rtype: integer"},
 
-  // interface to igraph_is_dag
+  // interface to igraph_is_dag`
```
ntamas: This is probably not needed :)

src/_igraph/graphobject.c
```diff
@@ -15699,6 +15828,34 @@ struct PyMethodDef igraphmodule_Graph_methods[] = {
    "  original implementation is used.\n"
    "@return: the community membership vector.\n"
   },
+  {"community_leiden",
+   (PyCFunction) igraphmodule_Graph_community_leiden,
+   METH_VARARGS | METH_KEYWORDS,
+   "community_leiden(edge_weights=None,node_weights=None\n"
```
ntamas: Can we add spaces after the commas in the docstring?

src/_igraph/igraphmodule.c
```diff
@@ -840,6 +840,9 @@ extern PyObject* igraphmodule_arpack_options_default;
   PyModule_AddIntConstant(m, "TRANSITIVITY_NAN", IGRAPH_TRANSITIVITY_NAN);
   PyModule_AddIntConstant(m, "TRANSITIVITY_ZERO", IGRAPH_TRANSITIVITY_ZERO);
 
+  PyModule_AddIntConstant(m, "CPM", 0);
```
ntamas: You might not even need these - just use strings on the Python side (i.e. `"cpm"` and `"modularity"`). This would be consistent with the enums in the C core where we also allow string representations in the Python layer - these are converted on-the-fly to the appropriate integers (see the conversion functions in `convert.c`).

src/_igraph/graphobject.c
vtraag: Haha, no, I don't think so :)

src/_igraph/igraphmodule.c
vtraag: I'm not entirely sure if I understand what you mean here. To be sure, the only place I actually use this constant is here:
https://github.com/igraph/python-igraph/blob/c6f6c1d01e0564077177d2cec5d49d5e8da0cf00/src/igraph/__init__.py#L1428-L1435
In other words, it is not a proper enum in the `igraph` C core, it is just something that I use in Python. Do you prefer to use strings instead?

ntamas: Yes, please, just use a single string instead if that's okay with you. I'd rather not add arbitrary integer aliases to the `igraph._igraph` module any more; the output of `dir(igraph._igraph)` is polluted enough as it is.

vtraag:
I've implemented the test. Pending you answer to the [enum issue](https://github.com/igraph/python-igraph/pull/273#discussion_r375265366), it is good to go, as far as I am concerned.

The only thing I was wondering was the random element in the unit tests. Due to random chance, it is possible (although unlikely) that a test yields a different outcome than what is expected. But this seems to be the case for many other tests as well. Is the RNG seeded anywhere for the tests?

ntamas:
Just seed the RNG in the test function itself. This is the only way to ensure determinism because the execution order of the tests is not guaranteed by the test runner.

FWIW, the Windows tests seem to be failing on Appveyor.

vtraag:
Tests are now made deterministic (I was actually also a bit too quick to judge that that a different outcome was unlikely), and I am now using a string instead of a constant.

vtraag:
By the way, I don't understand something: I see that the [build](https://travis-ci.org/igraph/python-igraph/builds/646586009) of this PR gets triggered on Travis, but I don't see the build status reported here in the PR. I'm not immediately sure how that can be resolved, but perhaps it would be good to check out why that is the case

ntamas:
I have no idea why this is the case; I tried removing and then reinstalling Travis in this repo, but it didn't help. I'll try to investigate it later.

iosonofabio:
My browser says: "This branch has no conflicts with the base branch". Is that the PR status you guys are missing?

ntamas:
No, we were missing the small yellow / red / green badges that show the result of the CI checks. But apparently it's fixed now - it started to appear for the last few commits.

iosonofabio:
Sounds good, yes the badges can be significantly delayed in my experience with github/Travis

llinfeng:
@vtraag How to seed `community_leiden`? I was pulling for a quotient graph [in this demo](https://github.com/llinfeng/pyhton-igraph-multigraph-demo/blob/master/Demo-QuotientGraph.ipynb) and the partitions returned from `community_leiden` are different every time I call it.

vtraag:
@llinfeng, please use the [igraph community forum](https://igraph.discourse.group/) for support questions.

You can set the random seed for the `igraph` in general by first [setting a random number generator](https://igraph.org/python/doc/igraph._igraph-module.html#set_random_number_generator) and seeding that.
```python
import random
import igraph as ig
ig.set_random_number_generator(random)
random.seed(0) # Or any other number of course
```

Issue #93, opened 2016-09-05
example python code for igraph


idc9:
I recently started using igraph and find the process of orienting myself to the package a little slow due to the lack of example code. A number of my collaborators had similar experiences. The [tutorial](http://igraph.org/python/doc/tutorial/tutorial.html) is certainly a good starting place and I've found a few other sources of example code. There is a lot of [example code](https://github.com/kolaczyk/sand) in R from the SAND textbook, however, not as many (that I can find) in Python. 

I find that small worked out examples (that display results of the code) are the best way to learn a new package. For example check out the examples of how to do basic tasks with pandas on [Chris Albon's website](http://chrisalbon.com/). These were far more helpful for learning how to use pandas than the pandas documentation.

**igraph is a wonderful package and deserves example python code that lowers the barrier to entry for new users**.

Does anyone know of a good source of example code that I'm missing? If so these should be linked to on the website. Otherwise it might be worth creating a bunch of examples and posting them to the website.

ntamas:
You are totally right, the documentation of python-igraph is a bit lacking, mostly due to lack of spare time on my side (I am not working in academia any more so I am unable to dedicate as much time to igraph as I used to). If anyone is willing to contribute example code, I'd be happy to link to it from the main webpage or even include it on the main webpage.

adodell:
@ntamas I'd be willing to redo the Tutorial on the python-igraph site, since it's written for Python 2.7 and igraph 6.0. It also looks like a few of the visuals are wrong in the tutorial.

I could even write some more documentation or examples, if you give me some direction on what is needed and the format you'd like the docs/examples to be written in.

ntamas:
Thanks for your offer! The tutorial is in the python-igraph repo in the `doc/source` subdirectory; the pages are written in [reStructuredText](http://docutils.sourceforge.net/rst.html) and the figures are plain PNG files (see the `doc/source/figures` subdir). Feel free to make any modifications in a fork and then send a pull request with the documentation improvements.

szhorvat:
@adodell Are you still feeling up for it? :-)

vtraag:
Regarding this, @jboynyc is working on this in PR #279.  Perhaps we should already merge that PR, so that we already include those contributions? We can then continue with improving the documentation.

jboynyc:
My available time to work on the documentation overhaul has shrunk pretty dramatically, so I think merging in incremental changes might be the best bet right now.

ntamas:
[...]

PR #279, opened 2020-02-13
Bring tutorial and documentation up to date.

jboynyc:
In response to the request in issue #93, I thought I would take a few stabs at updating and improving the tutorial. Here's a first set of edits.

### DIFF
--- a/doc/source/install.rst
+++ b/doc/source/install.rst
@@ -8,26 +8,24 @@
 Installing |igraph|
 ===================
 
-This chapter describes how to install the C core of |igraph| and its Python bindings
-on various operating systems.
+This chapter describes how to install the C core of |igraph| and its Python bindings on
+Windows, macOS, and GNU/Linux.
 
 Which |igraph| is right for you?
 ================================
 
-|igraph| is primarily a library written in C. It is *not* a standalone program, nor it is
-a Python package that you can just drop on your Python path to start using it. Therefore,
-if you would like to exploit |igraph|'s functionality in Python, you must install
-a few packages. Do not worry, though, there are precompiled packages for the major operating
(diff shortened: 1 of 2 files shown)


vtraag:
Hi @jboynyc! Thanks for the edits! To be sure: you are planning on committing more? If so, just let us know when you would like us to take a closer look. FYI: I assume you noticed that a new version of `python-igraph` was just released (version 0.8.0), including binaries from PyPI for all platforms? There may be some differences with earlier versions throughout the documentation.

For the next time, you can also mark your PR as a [draft PR ](https://github.blog/2019-02-14-introducing-draft-pull-requests/) (recently learned about that myself actually), but for now we can simply stick to this.

jboynyc:
Yes, I saw. Congrats @vtraag and team! Exciting times.

I will make more edits, also to some other parts of the documentation. Good idea to use a draft PR that I add commits to rather than making a bunch of separate PRs. (The feature was new to me, too.)

vtraag:
> Yes, I saw. Congrats @vtraag and team! Exciting times.

Thanks!

> I will make more edits, also to some other parts of the documentation. Good idea to use a draft PR that I add commits to rather than making a bunch of separate PRs. (The feature was new to me, too.)

OK, great to hear you will also be editing some other parts of the documentation! Note that you can also simply push new commits to this PR (as it also says here below): 

Add more commits by pushing to the `update-tutorial` branch on `jboynyc/python-igraph`.

jboynyc:
> Note that you can also simply push new commits to this PR (as it also says here below):

Oh right, that's what I'll do then. (I haven't used GitHub for collaboration in a while, so I forgot some of the ins and outs.)

PR #285, opened 2020-03-04
Sparse adjacency fix

### LINKED ISSUE #284: NameError from Graph.get_adjacency_sparse
The `Graph.get_adjacency_sparse` method seems to be broken in `v0.8.0`. Here's a small example:

```python
In [1]: import igraph                                                                                                          

In [2]: igraph.Graph.Erdos_Renyi(100, .3).get_adjacency_sparse()                                                               
NameError                                 Traceback (most recent call last)
<ipython-input-2-35bd1044fa8e> in <module>
----> 1 igraph.Graph.Erdos_Renyi(100, .3).get_adjacency_sparse()

/usr/local/lib/python3.7/site-packages/igraph/__init__.py in get_adjacency_sparse(self, attribute)
    620         if not self.is_directed():
    621             sparse_matrix = sparse_matrix + sparse_matrix.T
--> 622             di = np.diag_indices(len(edges))
    623             sparse_matrix[di] /= 2
    624         return sparse_matrix

NameError: name 'np' is not defined

In [3]: igraph.__version__                                                                                                     
Out[3]: '0.8.0'

```

ivirshup:
Fixes #284

### DIFF
--- a/tests/test_conversion.py
+++ b/tests/test_conversion.py
@@ -86,6 +86,25 @@ class GraphRepresentationTests(unittest.TestCase):
             [0, 0, 0, 0, 0, 0]
         ]))
 
+    def testGetSparseAdjacency(self):
+        # Undirected case
+        import numpy as np
+        g = Graph.Tree(6, 3)
+        g.es["weight"] = range(5)
+        self.assertTrue(np.all(
+            (g.get_adjacency_sparse() == np.array(g.get_adjacency().data))
+        ))
+        self.assertTrue(np.all(
+            (g.get_adjacency_sparse(attribute="weight") == np.array(g.get_adjacency(attribute="weight").data))
+        ))
+
+        # Directed case
+        g = Graph.Tree(6, 3, "tree_out")
+        g.add_edges([(0,1), (1,0)])
+        self.assertTrue(np.all(
+            g.get_adjacency_sparse() == np.array(g.get_adjacency().data)
+        ))
+
 
 def suite():
     direction_suite = unittest.makeSuite(DirectedUndirectedTests)
--- a/tests/test_conversion.py
+++ b/tests/test_conversion.py
@@ -91,7 +91,7 @@ class GraphRepresentationTests(unittest.TestCase):
             from scipy import sparse
             import numpy as np
         except ImportError:
-            self.skipTest("Scipy is a dependency of this test.")
+            self.skipTest("Scipy and numpy are dependencies of this test.")
 
         # Undirected case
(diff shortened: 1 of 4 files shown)


vtraag:
Thanks for raising the issue and providing a fix! 

The Travis CI build fails because currently neither `numpy` nor `scipy` are installed on the CI server. Neither package will be made a strict dependency of `igraph`, I think. We could install try to install `numpy` and `scipy` both in the `before_install` in `travis.yml`, it seems that they are now `pip` installable (that only seems to be possible for Python>=3.5, but Python 2.7 is phased out anyway). What do you think @ntamas? The test should probably be [skipped](https://docs.python.org/2/library/unittest.html#skipping-tests-and-expected-failures) if `numpy` or `scipy` is not installed by a user.

ivirshup:
No problem!

I'd added scipy to the travis dependencies in e33d06441434b7ebc92a6f82caf8bcafa8990ab9, but that doesn't seem to make it available during testing. I'm trying the `tox.ini` file now. I'm thinking I'll add the option to skip the test once I've see it run on travis.

ntamas:
`tox` creates a clean virtual environment for testing with only minimal dependencies and `python-igraph` itself. If we want to install `numpy` in CI, that's OK (I'd say it would be preferred anyway), we need to add it to the [`deps` option in `tox.ini`](https://tox.readthedocs.io/en/latest/config.html#conf-deps).

ivirshup:
Yep, looks like that's done it. I've just added the installation check. Let me know if there's anything else!

Should external dependencies be added to the wheels tests too? I was thinking no.

ntamas:
I think the `deps` line supports the full [PEP508](https://www.python.org/dev/peps/pep-0508/) specifier syntax, so you can probably use something like `platform_python_implementation !== 'PyPy`'. (You need to check what `platform.python_implementation()` returns in PyPy and exclude that).

tests/test_conversion.py
```diff
@@ -87,8 +87,13 @@ def testGetAdjacency(self):
         ]))
 
     def testGetSparseAdjacency(self):
+        try:
+            from scipy import sparse
+            import numpy as np
+        except ImportError:
+            self.skipTest("Scipy is a dependency of this test.")
```
vtraag: Probably better to mention both `numpy` and `scipy` in the skip message?

ivirshup:
Changed the skip message and fixed the build deps. Should I do anything else?

ntamas:
I think it's good to go. Thank you for your contribution!

PR #294, opened 2020-05-03
Extend get_subisomorphisms_vf2 to accept callback function that runs …

harenbergsd:
This PR is for letting **get_subisomorphisms_vf2**  accept a callback function.  See: [https://igraph.discourse.group/t/stop-subgraph-isomorphism-after-x-matches/185/4](https://igraph.discourse.group/t/stop-subgraph-isomorphism-after-x-matches/185/4). This allows you to do things like stop the algorithm if you have hit a number of matches.

### DIFF
--- a/src/_igraph/graphobject.c
+++ b/src/_igraph/graphobject.c
@@ -8477,6 +8477,7 @@ typedef struct {
   PyObject* callback_fn;
   PyObject* graph1;
   PyObject* graph2;
+  igraph_vector_ptr_t* result;
 } igraphmodule_i_Graph_isomorphic_vf2_callback_data_t;
 
 igraph_bool_t igraphmodule_i_Graph_isomorphic_vf2_callback_fn(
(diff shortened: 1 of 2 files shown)


harenbergsd:
Hmm perhaps adding a callback to get_subisomorphisms_vf2  has little benefit. I think it's possible to use subisomorphic_vf2, as you can pass a callback and store results like you said (and have it continue processing indefinitely).

It seems like get_subisomorphisms_vf2 is just a convenience wrapper around subisomorphic_vf2. Unlike subisomorphic_vf2, it returns all the results without having to make your own callback. So there is probably no reason to have get_subisomorphisms_vf2 accept a call back unless it does something like return all the results. Otherwise, it is effectively the same as subisomorphic_vf2, right?

ntamas:
Hmmm, you are right, I did not notice that the callback was already implemented for `subisomorphic_vf2()`. You are right, `get_subisomorphisms_vf2()` is simply a convenience wrapper and there's no point in adding the callback interface to `get_subisomorphisms_vf2()`.

I'm closing this PR; feel free to reopen if you disagree and you'd like to work on this further.

PR #297, opened 2020-05-14
setup.py: Support MinGW Windows build

Parakoopa:
Currently the setup.py tries to compile the Python extension using Visual Studio / MSVC.

However some workflows may use the MinGW version of Python instead (eg. when using GTK+ with Python). Building it is pretty simple, it can just be installed like on Linux and the Msys requirements are the same as for the build of the C core.

This pull request changes the check for build on Windows to only use MSVC when running the "native"/official CPython interpreter.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,6 +26,7 @@ import glob
 import shutil
 import subprocess
 import sys
+import sysconfig
 
 from select import select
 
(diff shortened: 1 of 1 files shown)


ntamas:
Thanks! I'll check this PR soon; it seems like there's something wrong with Appveyor right now so the build fails there, but it's probably unrelated to the PR.

I'll try to fix Appveyor first and then merge this if the build passes there.

ntamas:
Can you please merge [this commit](https://github.com/igraph/python-igraph/commit/b94ff802ece20f00107c4d17dcb0d72578e2b349) into your PR to fix the Appveyor CI build?

ntamas:
Thanks a lot and sorry for the delay!

Issue #229, opened 2019-07-03
Graph.Incidence should accept weighted argument


jboynyc:
- [X] This issue is for the Python interface of igraph.
- [X] This issue is a bug report or a feature request, not a support question.

The Python interface supports creating bipartite graphs from incidence matrices using [Graph.Incidence](https://igraph.org/python/doc/igraph.Graph-class.html#Incidence). The corresponding function `graph_from_incidence_matrix` in the R library accepts an argument `weighted` (see [documentation](https://igraph.org/r/doc/graph_from_incidence_matrix.html)) to treat non-zero matrix values as edge weights, while the Python library does not. 

It would be nice for python-igraph to have feature partity with R-igraph.

PuneethaPai:
Hi @ntamas , @jboynyc 
I would like to contribute to this. Any help would be appreciated. I have contributed to pure python libraries.
But this is my first time trying to write a wrapper on c-igraph.
Any help on where to start, how to debug would be great.

I know this is the place
```python
    @classmethod
    def Incidence(klass, *args, **kwds):
        result, types = klass._Incidence(*args, **kwds)
        result.vs["type"] = types
        return result
```
But as soon as i pass `weighted=True` i get invalid error, but couldn't find from where it is thrown.

Thanks in advance.

ntamas:
The `_Incidence` method is implemented in the glue code between Python and igraph's C core:

https://github.com/igraph/python-igraph/blob/master/src/_igraph/graphobject.c#L2528

As you can see, this function does not accept a `weighted` argument (neither as a positional nor as a keyword argument). You need to exclude that argument from `kwds` before calling `_Incidence` and then you need to take care of setting up the weights in the Python layer.

PuneethaPai:
Hi @ntamas 
Can you check the linked pr and share feedback.
Thanks
_PS: sorry for bothering :)_

ntamas:
I'll try to take care of it later today.

PR #303, opened 2020-06-21
Weighted incidence

PuneethaPai:
Fixes Issue: #229 

I have handled following cases w.r.t permutations of params:

- [x]  Allow empty str as weighted attr name
- [x]  Raise `ValueError` when multiple and weighted can not co-exist.
- [x] `weighted=True` graph becomes weighted with edge_attr=`weight`
- [x] `weighted="some_str"` graph is un-weighted with edge_attr=`some_str`
- [x] `directed=True, mode="out", weighted=True` graph is weighted, direction `out` is respcted
- [x] `directed=True, mode="in", weighted=True` graph is weighted, direction `in` is respcted
- [x] `directed=True, mode="all", weighted=True` graph is weighted, direction `all` is respcted

### DIFF
--- a/src/igraph/__init__.py
+++ b/src/igraph/__init__.py
@@ -2817,6 +2817,13 @@ class Graph(GraphBase):
           what the value is. If C{True}, non-zero entries are rounded up to
           the nearest integer and this will be the number of multiple edges
           created.
+        @param weighted: defines whether to create a weighted graph from the
+          incidence matrix. If it is c{None} then an unweighted graph is created
+          and the multiple argument is used to determine the edges of the graph.
+          If it is a character constant then for every non-zero matrix entry, an
+          edge is created and the value of the entry is added as an edge attribute
+          named by the weighted argument. If it is C{True} then a weighted graph
+          is created and the name of the edge attribute will be ‘weight’.
 
         @return: the graph with a binary vertex attribute named C{"type"} that
           stores the vertex classes.
@@ -2825,18 +2832,17 @@ class Graph(GraphBase):
         if weighted:
           kwds["multiple"] = False
         result, types = klass._Incidence(*args, **kwds)
+        result.vs["type"] = types
         if weighted:
-            mat = args[0]
             weight_attr = "weight" if weighted == True else weighted
-            result.es[weight_attr] = 1
+            mat = args[0]
+            _, rows, columns = result.get_incidence()
             for edge in result.es:
                 source, target = edge.tuple
-                if result.is_directed() and kwds.get("mode", "out") == "in":
-                    row, column = target, source - len(mat)
+                if source in rows:
+                    edge[weight_attr] = mat[source][target - len(rows)]
                 else:
-                    row, column = source, target - len(mat)
-                edge[weight_attr] = mat[row][column]
-        result.vs["type"] = types
+                    edge[weight_attr] = mat[target][source - len(rows)]
         return result
 
     def bipartite_projection(self, types="type", multiplicity=True, probe1=-1,
--- a/src/igraph/__init__.py
+++ b/src/igraph/__init__.py
@@ -2825,15 +2825,19 @@ class Graph(GraphBase):
           weighted argument. If it is C{True} then a weighted graph is created and
           the name of the edge attribute will be ‘weight’.
 
+        @raise ValueError: if the weighted and multiple are passed together.
+
         @return: the graph with a binary vertex attribute named C{"type"} that
           stores the vertex classes.
         """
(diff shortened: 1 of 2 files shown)


PuneethaPai:
Hi @ntamas ,
In test code I have used:
- `all(list([bool]))` instead of `bool and bool and ...`
- I have also used `self.assertListEquals(list1, list2)` instead of `self.assertTrue(list1 == list2)`

I did not modify other parts of code.

Now to maintain homogeneity:
- If you liked this I can apply the same to other parts of code.
- If you do not like this I can change it back to how you all are following.

PuneethaPai:
I think i haven't handled negative weight scenario. I will refer to behavior w.r.t R implementation and accordingly handle in python.
Let me know if you have any thoughts how to tackle negative weight.
Thanks

PuneethaPai:
@ntamas 
Just checked with R-igraph. It allows negative values to the weights.
So I think we are good. I was in assumption that we would not allow negative weights.
You can merge this. Let me know if there are any cases I have missed or any improvements in code.

Thanks

src/igraph/__init__.py
```diff
@@ -2814,12 +2814,32 @@ def Incidence(klass, *args, **kwds):
...
+          edge is created and the value of the entry is added as an edge attribute
+          named by the weighted argument. If it is C{True} then a weighted graph
+          is created and the name of the edge attribute will be ‘weight’.
 
         @return: the graph with a binary vertex attribute named C{"type"} that
           stores the vertex classes.
         """
+        weighted = kwds.pop("weighted", False)
+        if weighted:
+          kwds["multiple"] = False
```
ntamas, line 2830: If `multiple` is already given by the user as a keyword argument, don't overwrite it, raise an exception instead and ask the user to sort it out.

src/igraph/__init__.py
```diff
@@ -2814,12 +2814,32 @@ def Incidence(klass, *args, **kwds):
           what the value is. If C{True}, non-zero entries are rounded up to
           the nearest integer and this will be the number of multiple edges
           created.
+        @param weighted: defines whether to create a weighted graph from the
+          incidence matrix. If it is c{None} then an unweighted graph is created
+          and the multiple argument is used to determine the edges of the graph.
+          If it is a character constant then for every non-zero matrix entry, an
```
ntamas, line 2820: Does not need to be a "constant" - it can also be a variable that happens to be a string :) I'd simply write "string" instead of "character constant"

src/igraph/__init__.py
```diff
@@ -2814,12 +2814,32 @@ def Incidence(klass, *args, **kwds):
...
         @return: the graph with a binary vertex attribute named C{"type"} that
           stores the vertex classes.
         """
+        weighted = kwds.pop("weighted", False)
+        if weighted:
+          kwds["multiple"] = False
         result, types = klass._Incidence(*args, **kwds)
         result.vs["type"] = types
+        if weighted:
+            weight_attr = "weight" if weighted == True else weighted
```
ntamas, line 2834: We are not being entirely consistent here. If `weighted` is set to an empty string, we treat it as if it was `False`, contrary to what the documentation says.

I would do a more thorough type checking first on `weighted`; basically we need to ensure first that `weighted` can only be one of `True`, `False` or a string. Once we've established that, we need to treat the empty string as any other string.

src/igraph/__init__.py
```diff
@@ -2814,12 +2814,32 @@ def Incidence(klass, *args, **kwds):
...
+            mat = args[0]
+            _, rows, columns = result.get_incidence()
+            for edge in result.es:
+                source, target = edge.tuple
+                if source in rows:
+                    edge[weight_attr] = mat[source][target - len(rows)]
```
ntamas, line 2840: Lift `len(rows)` to a variable outside the loop (say `num_vertices_of_first_kind`) and just use that variable inside the loop to avoid calling `len()` all the time.

src/igraph/__init__.py
PuneethaPai: Can graph allow empty string property in edges?

ntamas: In theory it does; the attribute storage is simply a Python dict where the keys are enforced to be strings, but AFAIK there are no other constraints.

PuneethaPai:
@ntamas  why is wheel build is not happening in win32 x86 system? How to debug this issue?

ntamas:
This is probably an issue with the C core and is unrelated to your PR. I'll merge this and then fix the failing build in a separate commit. Thanks for your contribution!

Issue #29, opened 2015-05-02
Pandas interface for reading files


tcfuji:
Since the R-version of igraph makes reading files (for example, in csv format) easy, could the Python version use something similar? Something like `igraph.read_csv`?

ntamas:
You can read graphs in CSV format into igraph like this:

```
import csv, igraph
reader = csv.reader(open("input.csv"))
graph = igraph.Graph.TupleList(reader)
```

Or, if the CSV file contains headers:

```
reader = csv.DictReader(open("input.csv"))
graph = igraph.Graph.DictList(vertices=None, edges=reader)
```

Of course there is the possibility of wrapping these in a `Graph.Read_CSV` method, but there are too many possibilities with the interpretation of the columns in the CSV file (e.g., are there headers? which columns contain the endpoints of the edges? is the graph directed? which remaining columns should be stored as edge attributes? how can one provide attributes for the vertices? which vertex attribute should be used to store the names of the vertices?) that I believe it is hard to make a CSV reader that caters for all these use-cases, and it is yet unclear to me what is the "core" functionality that should be covered by the function. (The truth is, I don't really use CSV files to store my data). If you happen to have a good idea of what should be handled by `Graph.Read_CSV` and which of the above questions are not relevant for more than 20% of the use-cases, I'm happy to discuss this further.

bill10:
any progress? The solution from @ntamas works for reading in an edge list but does not work as general as the "graph_from_data_frame" counterpart in R. If there are isolates in the network then this solution won't be able to read in the isolates as they won't show up in the edgelist. Also, the "graph_from_data_frame" in R can read in node attributes and edge attributes easily.

szhorvat:
Does this help? https://igraph.discourse.group/t/how-to-create-a-graph-from-pandas-dataframes/34

iosonofabio:
We have now implemented ways of reading from pandas dataframes, so you can read the CSV into pandas and then come into igraph that way.

Since CSV is not a standardized graph format, reinventing the wheel and also designing new file formats seems like a questionable endeavour given the pandas support.

So I'm going to close this one.

ntamas:
@iosonofabio The CI build is failing with a recent version of Pandas, see [here](https://travis-ci.org/github/igraph/python-igraph/jobs/723406293). As I am not familiar with Pandas, can you please take a look at what's wrong there?

iosonofabio:
[...]

PR #309, opened 2020-07-14
Loading graph from one or two dataframes

iosonofabio:
As requested in #29, a function to read graphs from one/two dataframes, just like in R.

We are not going to implement the CSV parser anyway, so if this is fine (might need some tests?) then we can add the CSV solutions from @ntamas to the docs (I'll do that) and then close that issue.

### DIFF
--- a/src/igraph/__init__.py
+++ b/src/igraph/__init__.py
@@ -2894,7 +2894,7 @@ class Graph(GraphBase):
             if names.duplicated().any():
                 raise ValueError('Vertex names must be unique')
 
-            if np.setdiff1d(names_edges, names.values):
+            if len(np.setdiff1d(names_edges, names.values)):
                 raise ValueError(
                     'Some vertices in the edge DataFrame are missing from vertices DataFrame')
 
--- a/src/igraph/__init__.py
+++ b/src/igraph/__init__.py
@@ -2901,10 +2901,7 @@ class Graph(GraphBase):
             names = names.values
 
         # create graph
-        g = Graph(0, directed=directed)
-
-        # add vertices
-        g.add_vertices(len(names))
+        g = Graph(n=len(names), directed=directed)
 
         # vertex attributes
         if vertices is not None:
--- a/src/igraph/__init__.py
+++ b/src/igraph/__init__.py
@@ -2850,6 +2850,87 @@ class Graph(GraphBase):
                     edge[weight_attr] = mat[target][source - num_vertices_of_first_kind]
         return result
 
+    @classmethod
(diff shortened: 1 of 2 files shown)


tests/test_generators.py
```diff
@@ -164,6 +164,28 @@ def testWeightedAdjacency(self):
         self.assertTrue(el == [(0,1), (0,2), (1,0), (3,1)])
         self.assertTrue(g.es["w0"] == [1, 2, 2, 1])
 
+    def testDataFrame(self):
```
ntamas, line 167: Use the `skipIf()` decorator from `tests/utils.py` to skip the test case when `pandas` is not installed. See `tests/test_basic.py` for inspiration; there is a test case that is skipped when NumPy is not installed.

iosonofabio: :+1:

src/igraph/__init__.py
```diff
@@ -2850,6 +2850,84 @@ def Incidence(klass, *args, **kwds):
...
+        """
+        import numpy as np
+        import pandas as pd
+
+        if edges.shape[1] < 2:
+            raise ValueError("the data frame should contain at least two columns")
+
+        # Handle if some elements are 'NA'
+        if edges.iloc[:, :2].isna().values.any():
+            warn("In `d' `NA' elements were replaced with string \"NA\"")
```
ntamas, line 2876: what is `d'?

src/igraph/__init__.py
```diff
@@ -2850,6 +2850,84 @@ def Incidence(klass, *args, **kwds):
...
+            edges = edges.copy()
+            edges.iloc[:, :2].fillna('NA', inplace=True)
+
+        if (vertices is not None) and vertices.iloc[:, 0].isna().values.any():
+            warn("In `vertices.iloc[:, 0]` `NA' elements were replaced with string \"NA\"")
```
ntamas, line 2881: I'm not that familiar with Pandas to judge this -- would "vertices.iloc[:, 0]" make sense for an average user that knows Pandas but has no idea about how this function is implemented internally in igraph? Can we come up with a more readable warning?

src/igraph/__init__.py
iosonofabio: a bad copycat from R ;-) correcting now

src/igraph/__init__.py
iosonofabio: fair enough, yes I put a more readable warning now

iosonofabio:
I think this can be merged now

Issue #269, opened 2015-09-23
To implement graph union and graph difference by vertex names in Python, in the same way it is already implemented in R


cndesantana:
The native C core of igraph supports graph union only by IDs and not
by vertex names. The R interface contains an implementation of "graph
union by vertex names", and it is selected automatically if your graph
has vertex names. However, There is no equivalent function in the Python
interface.

It would be great if you could implement graph union and graph difference by vertex names in Python, in the same way it is already implemented in R.

This issue was also discussed here: https://lists.nongnu.org/archive/html/igraph-help/2015-09/msg00083.html

Thanks

ntamas:
No, it's not solved yet; there are too many corner cases to consider (see my [earlier comment](https://github.com/igraph/python-igraph/issues/269#issuecomment-263817279) and I haven't had time to implement it properly.

ntamas:
Implemented in #310 so closing.

Issue #82, opened 2016-06-07
union does not keep attributes


rokroskar:
apologies if this is the wrong place for such a question, but: it appears that currently union does not keep attributes? I found an [old thread](http://comments.gmane.org/gmane.comp.science.graph.igraph.general/1541) which seemed to suggest that an option `keep_attributes` would be added, but I can't find such an option anywhere. Is there a different method that should be used instead?

ntamas:
True; `union` does not keep attributes and I haven't managed to implement this properly so far due to lack of time. There is no alternative implementation yet, so you'll need to code the `union` operator yourself if you want to keep the attributes. If you manage to come up with an implementation, please send us a pull request and then I'll try to integrate it into the master branch.

rokroskar:
Ok thanks, just wanted to make sure I wasn't reinventing the wheel. Was the plan to implement it on the C library side or in python? My workaround in python is rather naive (keeping lists of weights and combining them in the end) and it will get very slow for large graphs I imagine.

ntamas:
The original plan was to implement it in C because then all the higher level interfaces would get the functionality "for free". But of course it's much more complicated to do in C than in Python or R.

rodrigolece:
+1 for this feature request!

I was recently trying to create hierarchical weighted graphs and my first idea was to used the overloaded addition or multiplication of graphs, but since the `disjoint_union` method does not copy the attributes it couldn't be done in this way.

iosonofabio:
#310 deals only with union/intersection, not disjoint union, however it seems easy enough to include that one as well. @ntamas?

ntamas:
I'd be very grateful if you could add disjoint union as well. I promise to get it reviewed as soon as possible; I have a very long backlog right now but eventually I'll get to it. Thanks a lot!

iosonofabio:
No worries mate, I'll add disjoint union. Let me figure out what it is

PR #310, opened 2020-07-15
Union/intersection of graphs with vertex names

iosonofabio:
Trying to address #82 and #269 

For now there is a stub of union and one for intersection. Much of the logic (but not all) is in place, details and tests missing.

The obvious question from my side at this stage is about architecture:

1. Is the file `operators.py` an ok place for this?
2. Is it ok to have these two global functions within the `igraph` namespace and then hook a call to them in (to be defined) instance methods of the `Graph` class (i.e. to-be `Graph.union`, `Graph.intersection`)?
3. If 2 is fine, then we can reroute the `__and__` and `__or__` operators into this code path

Thanks!
Fabio

### DIFF
--- a/src/_igraph/graphobject.c
+++ b/src/_igraph/graphobject.c
@@ -9562,13 +9562,13 @@ PyObject *igraphmodule_Graph_disjoint_union(igraphmodule_GraphObject * self,
  * \brief Creates the union of two or more graphs
  */
 PyObject *igraphmodule_Graph_union(igraphmodule_GraphObject * self,
-                                   PyObject * args, PyObject * kwds,
-                                   )
+                                   PyObject * args, PyObject * kwds)
 {
   static char* kwlist[] = { "edgemaps", NULL };
   PyObject *it, *other;
   PyObject *with_edgemaps = Py_False;
-  igraphmodule_GraphObject *o, *result;
+  igraphmodule_GraphObject *o;
+  PyObject *result;
   igraph_t g;
 
   if (!PyArg_ParseTupleAndKeywords(args, kwds, "O|O", kwlist, &other, &with_edgemaps))
@@ -9599,7 +9599,8 @@ PyObject *igraphmodule_Graph_union(igraphmodule_GraphObject * self,
     /* prepare edgemaps if requested */
     if (PyObject_IsTrue(with_edgemaps)) {
       if (igraph_vector_ptr_init(edgemaps, 0)) {
-      return igraphmodule_handle_igraph_error();
+        return igraphmodule_handle_igraph_error();
+      }
     }
 
     /* Create union */
@@ -9622,7 +9623,7 @@ PyObject *igraphmodule_Graph_union(igraphmodule_GraphObject * self,
       Py_INCREF(str_edgemaps);
       for (i = 0; i < no_of_graphs; i++) {
         long int j;
-        long int no_of_edges = (long int) igraph_ecount(VECTOR(*gs)[i]);
+        long int no_of_edges = (long int) igraph_ecount(VECTOR(gs)[i]);
         PyObject *emi = PyList_New((Py_ssize_t) no_of_edges);
         Py_INCREF(emi);
         for (j = 0; j < no_of_edges; j++) {
@@ -9639,14 +9640,15 @@ PyObject *igraphmodule_Graph_union(igraphmodule_GraphObject * self,
       /* wrap in a dictionary */
       result = PyDict_New();
       Py_INCREF(result);
-      PyDict_SetItem(result, str_graph, o);
+      PyDict_SetItem(result, str_graph, (PyObject *) o);
       PyDict_SetItem(result, str_edgemaps, em_list);
 
       igraph_vector_ptr_destroy(edgemaps);
 
     }
     else {
-      CREATE_GRAPH(result, g);
+      CREATE_GRAPH(o, g);
+      result = (PyObject *) o;
     }
   }
   /* Did we receive only two graphs? */
@@ -9667,7 +9669,8 @@ PyObject *igraphmodule_Graph_union(igraphmodule_GraphObject * self,
     /* this is correct as long as attributes are not copied by the
      * operator. if they are copied, the initialization should not empty
      * the attribute hashes */
-    CREATE_GRAPH(result, g);
+    CREATE_GRAPH(o, g);
+    result = (PyObject *) o;
   }
 
   return (PyObject *) result;
@@ -9677,13 +9680,13 @@ PyObject *igraphmodule_Graph_union(igraphmodule_GraphObject * self,
  * \brief Creates the intersection of two or more graphs
  */
 PyObject *igraphmodule_Graph_intersection(igraphmodule_GraphObject * self,
-                                   PyObject * args, PyObject * kwds,
-                                   )
+                                   PyObject * args, PyObject * kwds)
 {
   static char* kwlist[] = { "edgemaps", NULL };
-  PyObject *it;
+  PyObject *it, *other;
   PyObject *with_edgemaps = Py_False;
-  igraphmodule_GraphObject *o, *result;
+  igraphmodule_GraphObject *o;
+  PyObject *result;
   igraph_t g;
 
   if (!PyArg_ParseTupleAndKeywords(args, kwds, "O|O", kwlist, &other, &with_edgemaps))
@@ -9714,7 +9717,8 @@ PyObject *igraphmodule_Graph_intersection(igraphmodule_GraphObject * self,
     /* prepare edgemaps if requested */
     if (PyObject_IsTrue(with_edgemaps)) {
       if (igraph_vector_ptr_init(edgemaps, 0)) {
-      return igraphmodule_handle_igraph_error();
+        return igraphmodule_handle_igraph_error();
+      }
     }
 
     /* Create intersection */
@@ -9737,7 +9741,7 @@ PyObject *igraphmodule_Graph_intersection(igraphmodule_GraphObject * self,
       Py_INCREF(str_edgemaps);
       for (i = 0; i < no_of_graphs; i++) {
         long int j;
-        long int no_of_edges = (long int) igraph_ecount(VECTOR(*gs)[i]);
+        long int no_of_edges = (long int) igraph_ecount(VECTOR(gs)[i]);
         PyObject *emi = PyList_New((Py_ssize_t) no_of_edges);
         Py_INCREF(emi);
         for (j = 0; j < no_of_edges; j++) {
@@ -9754,13 +9758,14 @@ PyObject *igraphmodule_Graph_intersection(igraphmodule_GraphObject * self,
       /* wrap in a dictionary */
       result = PyDict_New();
       Py_INCREF(result);
-      PyDict_SetItem(result, str_graph, o);
+      PyDict_SetItem(result, str_graph, (PyObject *) o);
       PyDict_SetItem(result, str_edgemaps, em_list);
 
       igraph_vector_ptr_destroy(edgemaps);
     }
     else {
-      CREATE_GRAPH(result, g);
+      CREATE_GRAPH(o, g);
+      result = (PyObject *) o;
     }
   }
   /* Did we receive only two graphs? */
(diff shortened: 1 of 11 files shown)


iosonofabio:
I now drafted a change in the `Graph.union` interface that recovers edgemaps from the C layer, just like in R. That way we should be able to keep edge attributes during union/intersection.

Because I'm a little rusty in Python C API calls, this might take a few iterations. Any comment on the interface would be welcome.

iosonofabio:
This is ready for review now. The changes reach quite a bit deeper than initially suspected. Summary:

- `operators.py` implements union and intersection of graphs like in the R interface - merging by name and keeping attributes
- `__init__.py` has changes in the `Graph` class to use those functions
- `operators.c/h` include modified interface functions dealing with edgemaps
- `graphobject.c/h` lost the original interfaces instead
- more tests are added to make sure the new functions work as expected

In addition, one trick was necessary, which is to add a modified conversion function that fills a vector of pointers to graphs and also stores the python type/class of the initial pointer (graph). That seems necessary to ensure the resulting union/intersection graph has the properties `vs` and `es` when generated at the C API interface.

One difference from R is that here the code is a little verbose but less obscure, whereas R uses a utility function to deal with attributes that shortens the code but makes it significantly harder to read.

iosonofabio:
Alright, I wrote the disjoint union and then lost it in a git stash... so I ended up recoding it.

Bottomline is now it's there and tested. Please have a look whenever you have a minute and we should be close to merging.

ntamas:
Thanks! This is a larger chunk so it will take me a while, stay tuned.

iosonofabio:
@ntamas I realize you are busy with other stuff, so if you would like me to do anything on this PR please let me know

src/_igraph/igraphmodule.c
```diff
@@ -650,14 +651,26 @@ static PyMethodDef igraphmodule_methods[] =
...
+  },
   {NULL, NULL, 0, NULL}
 };
 
 #define MODULE_DOCS \
   "Low-level Python interface for the igraph library. " \
   "Should not be used directly.\n\n"                    \
   "@undocumented: community_to_membership, _compare_communities, _power_law_fit, " \
-  "_split_join_distance"
+  "_split_join_distance, _union, _intersection"
```
ntamas, line 673: _disjoint_union is missing from here

src/_igraph/operators.c
```diff
@@ -0,0 +1,317 @@
...
+
+  /* Needs to be an iterable */
+  it = PyObject_GetIter(graphs);
+  if (!it) {
+      Py_DECREF(it);
+      return igraphmodule_handle_igraph_error();
+  }
+
+  /* Get all elements, store the graphs in an igraph_vector_ptr */
+  igraph_vector_ptr_t gs;
```
ntamas, line 55: Please move this declaration to the top of the function; ancient versions of MSVC will complain if you leave it here.

src/_igraph/operators.c
```diff
@@ -0,0 +1,317 @@
...
```
ntamas, line 123: Please move this declaration to the top of the function; ancient versions of MSVC will complain if you leave it here.

src/_igraph/operators.c
```diff
@@ -0,0 +1,317 @@
...
+    long int i;
+    em_list = PyList_New((Py_ssize_t) no_of_graphs);
+    for (i = 0; i < no_of_graphs; i++) {
+      long int j;
+      long int no_of_edges = (long int) igraph_ecount(VECTOR(gs)[i]);
+      igraph_vector_t *map = VECTOR(edgemaps)[i];
+      PyObject *emi = PyList_New((Py_ssize_t) no_of_edges);
+      for (j = 0; j < no_of_edges; j++) {
+        PyObject *dest = PyLong_FromLong(VECTOR(*map)[j]);
+        PyList_SetItem(emi, (Py_ssize_t) j, dest);
```
ntamas, line 161: You can use `PyList_SET_ITEM()` when initializing newly created lists; it is slightly faster than `PyList_SetItem`. `PyList_SET_ITEM()` will automatically steal a reference to `dest`.

If you don't use `PyList_SET_ITEM()` and stick to `PyList_SetItem`, then you will need to drop your own reference to `dest` because `PyList_SetItem` calls `Py_INCREF()` on `dest`.

src/_igraph/operators.c
```diff
@@ -0,0 +1,317 @@
...
+      }
+      PyList_SetItem(em_list, (Py_ssize_t) i, emi);
```
ntamas, line 163: Same here; you should either use `PyList_SET_ITEM` or drop your reference to `emi`.

src/_igraph/operators.c
```diff
@@ -0,0 +1,317 @@
...
+  else {
+    o = (igraphmodule_GraphObject*) igraphmodule_Graph_from_igraph_t(&g);
+  }
+
+  if (with_edgemaps) {
+    /* wrap in a dictionary */
+    PyObject * str_graph = PyUnicode_FromString("graph");
+    PyObject * str_edgemaps = PyUnicode_FromString("edgemaps");
+    result = PyDict_New();
+    PyDict_SetItem(result, str_graph, (PyObject *) o);
```
ntamas, line 195: You need to call `Py_DECREF(o)` here because `PyDict_SetItem` calls `Py_INCREF(o)` (it does not steal a reference).

Also, you can get rid of `str_graph` and `str_edgemaps` by using `PyDict_SetItemString`, which takes a `const char*` instead of a Python string object as the key.

src/_igraph/operators.c
```diff
@@ -0,0 +1,317 @@
...
+    long int i;
+    em_list = PyList_New((Py_ssize_t) no_of_graphs);
+    for (i = 0; i < no_of_graphs; i++) {
+      long int j;
+      long int no_of_edges = (long int) igraph_ecount(VECTOR(gs)[i]);
+      igraph_vector_t *map = VECTOR(edgemaps)[i];
+      PyObject *emi = PyList_New((Py_ssize_t) no_of_edges);
+      for (j = 0; j < no_of_edges; j++) {
+        PyObject *dest = PyLong_FromLong(VECTOR(*map)[j]);
+        PyList_SetItem(emi, (Py_ssize_t) j, dest);
```
ntamas, line 272: The usual comment about `PyList_SetItem` vs `PyList_SET_ITEM` and reference counts :)

src/_igraph/operators.c
```diff
@@ -0,0 +1,317 @@
...
+  else {
+    o = (igraphmodule_GraphObject*) igraphmodule_Graph_from_igraph_t(&g);
+  }
+
+  if (with_edgemaps) {
+    /* wrap in a dictionary */
+    PyObject * str_graph = PyUnicode_FromString("graph");
+    PyObject * str_edgemaps = PyUnicode_FromString("edgemaps");
+    result = PyDict_New();
+    PyDict_SetItem(result, str_graph, (PyObject *) o);
```
ntamas, line 307: The usual comment about `PyDict_SetItem` vs `PyDict_SetItemString` :) Also, you need to `Py_DECREF(o)` and `Py_DECREF(em_list)`.

src/_igraph/graphobject.c
```diff
@@ -15997,9 +15796,9 @@ PyNumberMethods igraphmodule_Graph_as_number = {
   (unaryfunc) igraphmodule_Graph_complementer_op, /*nb_invert */
   0,                            /*nb_lshift */
   0,                            /*nb_rshift */
-  (binaryfunc) igraphmodule_Graph_intersection, /*nb_and */
+  0,                            /*nb_and */
```
ntamas, line 15799: This row used to add support for using the `&` operator in Python between two graph objects to call the graph intersection function. I'm not 100% sure whether it was a good idea, but in order to keep the API stable, we must keep on supporting it. I don't mind if it is not supported any more by the low level `_GraphBase` object (which was never part of the public API), but we should re-add support for the `&` operator (in pure Python if that's easier) in the `Graph` object if we remove it from here.

src/_igraph/graphobject.c
```diff
@@ -15997,9 +15796,9 @@ PyNumberMethods igraphmodule_Graph_as_number = {
...
   0,                            /*nb_xor */
-  (binaryfunc) igraphmodule_Graph_union,  /*nb_or */
+  0,                            /*nb_or */
```
ntamas, line 15801: This row used to add support for using the `|` operator in Python between two graph objects to call the graph union function. I'm not 100% sure whether it was a good idea, but in order to keep the API stable, we must keep on supporting it. I don't mind if it is not supported any more by the low level `_GraphBase` object (which was never part of the public API), but we should re-add support for the `|` operator (in pure Python if that's easier) in the `Graph` object if we remove it from here.

ntamas (changes requested):
I finally managed to review your changes -- sorry it took such long time.

I'm mostly okay with the changes, but I have found a few Python reference leaks in the newly created `_disjoint_union()`, `_union()` and `_intersection()` functions that should definitely be fixed.

As for the attribute logic in `operators.py`, I feel it's okay and we are probably covered by the test cases, but I found the code hard to read due to the usage of many two- or three-letter variable names. If this PR is to be merged, I will probably go through these functions with a refactor tool and rename the variables to spell out their purpose explicitly.

src/_igraph/igraphmodule.c
iosonofabio: ok, added

src/_igraph/operators.c
iosonofabio: yep

src/_igraph/operators.c
iosonofabio: yep

src/_igraph/operators.c
iosonofabio: yep, thanks for the tip!

src/_igraph/operators.c
iosonofabio: yes yes good catch

src/_igraph/operators.c
iosonofabio: awesome improvment, of course there was a way

src/_igraph/operators.c
iosonofabio: yep fixed

src/_igraph/operators.c
iosonofabio: yes

src/_igraph/graphobject.c
iosonofabio: I added this in pure Python, and it is tested for. I tried several approaches at different stages of the interface, and at the end decided to just leverage the pure Python version because we need it anyway for the label stuff.

src/_igraph/graphobject.c
iosonofabio: As above for the &

iosonofabio:
Fantastic thank you for the thorough review!

I've fixed the things you mentioned. I'm in favour of refactoring if that helps. I've tried to rename the worst offenders, but there might be a few more that you still find confusing, so just go ahead and rename them into whatever seems to make sense.

Waiting for CI as I write.

iosonofabio:
I cleaned up the tests a little, including unrelated things that are minor. I know I should have done it in a separate PR and then rebase, but that seemed like a huge effort for such small things.

The takeaway is that the tests fail on my machine on a single thing now, which has to do with https://github.com/igraph/igraph/issues/597. That issue says you were going to rewire `python-igraph` to the fixed C core, has that happened already?

Thanks!

iosonofabio:
Well seems like that, since CI passes.

So it's done except for some optional refactoring of `operators.py`, at your discretion.

src/igraph/operators.py
```diff
@@ -0,0 +1,482 @@
...
+    @param graph: list of graphs. A lazy sequence is not acceptable.
+    @return: the disjoint union graph
+    """
+    if any(not isinstance(g, GraphBase) for g in graphs):
+        raise TypeError('Not all elements are graphs')
+
+    ngr = len(graphs)
+    # Trivial cases
+    if ngr == 0:
+        return None
```
ntamas, line 63: I'd rather throw an exception here. My first instinct would have been to return an empty graph, but we cannot decide whether it should be directed or undirected. An exception makes it clear that this is an ambiguous situation; furthermore, the `min()` and `max()` functions in Python also throw an exception for empty sequences so we are consistent with the rest of Python if we do it this way.

I'll take care of this today so we can merge this PR before the day starts for you :)

ntamas (approved):

iosonofabio:
Haha awesome, thank you!

PR #314, opened 2020-07-29
Docs on graph analysis

iosonofabio:
As done previously for graph generation, here a PR adding docs on graph analysis.

This one is a little less obvious because "graph analysis" can mean so many things. Vice versa, it helps us identify obvious holes in the Python interface such as DFS.

Please let me know what topics you would like me to write on.

Cheers,
Fabio

### DIFF
--- a/doc/source/analysis.rst
+++ b/doc/source/analysis.rst
@@ -2,7 +2,7 @@ Graph analysis
 ==============
 |igraph| enables analysis of graphs/networks from simple operations such as adding and removing nodes to complex theoretical constructs such as community detection. Read the `API documentation`_ for details on each function and class.
 
-The context for the following examples will be to import the :class:`Graph` class and to have one or more graphs available:
+The context for the following examples will be to import |igraph| (commonly as `ig`), have the :class:`Graph` class and to have one or more graphs available:
 
 >>> import igraph as ig
 >>> from igraph import Graph
@@ -14,7 +14,8 @@ To get a summary representation of the graph, use :meth:`Graph.summary`. For ins
 
 will provide a fairly detailed description.
 
-To copy a graph, use :meth:`Graph.copy`.
+To copy a graph, use :meth:`Graph.copy`. This is a "shallow" copy: any mutable objects in the attributes are not copied (they would refer to the same instance).
+If you want to copy a graph including all its attributes, use Python's `deepcopy` module.
 
 Vertices and edges
 +++++++++++++++++++++++++++
@@ -37,7 +38,13 @@ Similarly for edges, use :attr:`Graph.es`:
 >>> for e in g.es:
 >>>     print(e)
 
-.. note:: These two attributes are special sequences with their own useful methods. Some are described below, and all can be found in the `API documentation`_.
+You can index and slice vertices and edges similar to a list:
+
+>>> g.vs[:4]
+>>> g.vs[0, 2, 4]
+>>> g.es[3]
+
+.. note:: The `vs` and `es` attributes are special sequences with their own useful methods. See `API documentation`_ for a full list.
 
 If you prefer a vanilla edge list, you can use :meth:`Graph.get_edge_list`.
 
@@ -57,7 +64,7 @@ To get the edges incident on a vertex, you can use :meth:`Vertex.incident`, :met
 >>> v = g.vs[0]
 >>> edges = v.incident()
 
-The :meth:`Graph.adjecent` and :meth:`Graph.incident` functions fulfill the same purpose with a slightly different syntax based on vertex ids:
+The :meth:`Graph.incident` function fulfills the same purpose with a slightly different syntax based on vertex ids:
 
 >>> edges = g.incident(0)
 
@@ -71,7 +78,7 @@ To compute the neighbors, successors, and predecessors, the methods :meth:`Graph
 >>> neis = g.vs[0].neighbors()
 >>> neis = g.neighbors(0)
 
-To get the list of vertives within a certain distance of one or more initial nodes, you can use :meth:`Graph.neighborhood`:
+To get the list of vertices within a certain distance from one or more initial vertices, you can use :meth:`Graph.neighborhood`:
 
 >>> g.neighborhood([0, 1], order=2)
 
@@ -100,7 +107,6 @@ This changes the graph `g` in place. You can specify the name of the vertices if
 
 To remove nodes, use :meth:`Graph.delete_vertices`:
 
->>> g.delete_vertices(None)  # remove all vertices
 >>> g.delete_vertices([1, 2])
 
 Again, you can specify the names or the actual :class:`Vertex` objects instead.
@@ -112,8 +118,7 @@ To add edges, use :meth:`Graph.add_edge` and :meth:`Graph.add_edges`:
 
 To remove edges, use :meth:`Graph.delete_edges`:
 
->>> g.delete_edges(None)   # remove all edges
->>> g.delete_edges([0, 5]) # remove by edge ID
+>>> g.delete_edges([0, 5]) # remove by edge id
 
 You can also remove edges between source and target nodes.
 
@@ -153,14 +158,17 @@ To sort a graph topologically, use :meth:`Graph.topological_sorting`:
 >>> g = ig.Graph.Tree(10, 2, mode=ig.TREE_OUT)
 >>> g.topological_sorting()
 
-Graph traversing
+Graph traversal
 +++++++++++++++++++++
-A common operation is traversing the graph. |igraph| currently exposes breath-first search (BFS) via :meth:`Graph.bfs` and :meth:`Graph.bfsiter`:
+A common operation is traversing the graph. |igraph| currently exposes breadth-first search (BFS) via :meth:`Graph.bfs` and :meth:`Graph.bfsiter`:
 
 >>> [vertices, layers, parents] = g.bfs()
 >>> it = g.bfsiter()  # Lazy version
 
-A depth-first search function is in the works.
+Depth-first search has a similar infrastructure via :meth:`Graph.dfs` and :meth:`Graph.dfsiter`:
+
+>>> [vertices, parents] = g.dfs()
+>>> it = g.dfsiter()  # Lazy version
 
 To perform a random walk from a certain vertex, use :meth:`Graph.random_walk`:
 
@@ -182,6 +190,10 @@ As well as functions related to cuts and paths:
 - :meth:`Graph.mincut_value` - as previous one, but returns only the value
 - :meth:`Graph.all_st_cuts`
 - :meth:`Graph.all_st_mincuts`
+- :meth:`Graph.edge_connectivity` or :meth:`Graph.edge_disjoint_paths` or :meth:`Graph.adhesion`
+- :meth:`Graph.vertex_connectivity` or :meth:`Graph.cohesion`
+
+See also the section on flow.
 
 Global properties
 +++++++++++++++++++++
@@ -203,7 +215,6 @@ Connectedness:
 
 - :meth:`Graph.all_minimal_st_separators`
 - :meth:`Graph.minimum_size_separators`
-- :meth:`Graph.feedback_arc_set`
 - :meth:`Graph.cut_vertices` or :meth:`Graph.articulation_points`
 
 Cliques and motifs:
@@ -213,12 +224,13 @@ Cliques and motifs:
 - :meth:`Graph.maximal_cliques`
 - :meth:`Graph.largest_cliques`
 - :meth:`Graph.motifs_randesu` and :meth:`Graph.motifs_randesu_estimate`
-- :meth:`Graph.g.motifs_randesu_no` counts the number of motifs
+- :meth:`Graph.motifs_randesu_no` counts the number of motifs
 
-Structural:
+Directed acyclic graphs:
 
-- :meth:`Graph.edge_connectivity` or :meth:`Graph.edge_disjoint_paths` or :meth:`Graph.adhesion`
-- :meth:`Graph.vertex_connectivity` or :meth:`Graph.cohesion`
+- :meth:`Graph.is_dag`
+- :meth:`Graph.feedback_arc_set`
+- :meth:`Graph.topological_sorting`
 
 Optimality:
 
@@ -231,6 +243,7 @@ Optimality:
 - :meth:`Graph.largest_independent_vertex_sets`
 - :meth:`Graph.mincut`
 - :meth:`Graph.mincut_value`
+- :meth:`Graph.feedback_arc_set`
 - :meth:`Graph.maximum_bipartite_matching` (bipartite graphs)
 
 Other complex measures are:
@@ -238,16 +251,14 @@ Other complex measures are:
 - :meth:`Graph.assortativity`
 - :meth:`Graph.assortativity_degree`
 - :meth:`Graph.assortativity_nominal`
-- :meth:`Graph.coreness` (aka :meth:`Graph.shell_index`)
 - :meth:`Graph.density`
-- :meth:`Graph.hub_score`
 - :meth:`Graph.transitivity_undirected`
-- :meth:`Graph.transitivity_local_undirected`
 - :meth:`Graph.transitivity_avglocal_undirected`
 - :meth:`Graph.dyad_census`
 - :meth:`Graph.triad_census`
 - :meth:`Graph.reciprocity` (directed graphs)
 - :meth:`Graph.isoclass` (only 3 or 4 vertices)
+- :meth:`Graph.biconnected_components` aka :meth:`Graph.blocks`
 
 Boolean properties:
 
@@ -272,24 +283,25 @@ A spectrum of vertex-level properties can be computed. Similarity measures inclu
 Structural:
 
 - :meth:`Graph.authority_score`
+- :meth:`Graph.hub_score`
 - :meth:`Graph.betweenness`
 - :meth:`Graph.bibcoupling`
 - :meth:`Graph.closeness`
 - :meth:`Graph.constraint`
 - :meth:`Graph.cocitation`
+- :meth:`Graph.coreness` (aka :meth:`Graph.shell_index`)
 - :meth:`Graph.eccentricity`
-- :meth:`Graph.eigenvector_centrality` aka :meth:`Graph.evcent`
+- :meth:`Graph.eigenvector_centrality`
 - :meth:`Graph.pagerank`
 - :meth:`Graph.personalized_pagerank`
 - :meth:`Graph.strength`
+- :meth:`Graph.transitivity_local_undirected`
 
 Connectedness:
 
 - :meth:`Graph.subcomponent`
 - :meth:`Graph.is_separator`
 - :meth:`Graph.is_minimal_separator`
-- :meth:`Graph.biconnected_components`
-- :meth:`Graph.blocks`
 
 Edge properties
 +++++++++++++++
@@ -309,14 +321,14 @@ Matrix representations
 Matrix-related functionality includes:
 
 - :meth:`Graph.get_adjacency`
-- :meth:`Graph.get_adjacency_sparse` (sparase CSR matrix version)
+- :meth:`Graph.get_adjacency_sparse` (sparse CSR matrix version)
 - :meth:`Graph.laplacian`
 
 Clustering
 ++++++++++
 |igraph| includes several approaches to unsupervised graph clustering and community detection:
 
-- :meth:`Graph.clusters` (aka :meth:`Graph.components`): the connected components
+- :meth:`Graph.components` (aka :meth:`Graph.clusters`): the connected components
 - :meth:`Graph.cohesive_blocks`
 - :meth:`Graph.community_edge_betweenness`
 - :meth:`Graph.community_fastgreedy`
@@ -332,6 +344,10 @@ Clustering
 
 Simplification, permutations and rewiring
 +++++++++++++++++++++++++++++++++++++++++
+To check is a graph is simple, you can use :meth:`Graph.is_simple`.
+
+>>> g.is_simple()
+
 To simplify a graph (remove multiedges and loops), use :meth:`Graph.simplify`:
 
 >>> g_simple = g.simplify()
@@ -345,16 +361,20 @@ To permute the order of vertices, you can use :meth:`Graph.permute_vertices`:
 
 The canonical permutation can be obtained via :meth:`Graph.canonical_permutation`, which can then be directly passed to the function above.
 
-To rewire the graph at random while keeping some structural properties, there are:
+To rewire the graph at random, there are:
 
-- :meth:`Graph.rewire`
-- :meth:`Graph.rewire_edges`
+- :meth:`Graph.rewire` - preserves the degree distribution
+- :meth:`Graph.rewire_edges` - fixed rewiring probability for each endpoint
 
 Line graph
 ++++++++++
-To compute the line graph, there is :meth:`Graph.linegraph`:
+To compute the line graph of a graph `g`, which represents the connectedness of the *edges* of g, you can use :meth:`Graph.linegraph`:
 
+>>> g = Graph(n=4, edges=[[0, 1], [0, 2]])
 >>> gl = g.linegraph()
+
+In this case, the line graph has two vertices, representing the two edges of the original graph, and one edge, representing the point where
+those two original edges touch.
 
 Composition and subgraphs
 ++++++++++++++++++++++++++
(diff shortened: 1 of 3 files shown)


iosonofabio:
Alright, this is a good first stub. Of course, documentation can always get better, but I feel like at least now we have a compass for users to start using the library!

szhorvat:
Response to the review request: I can comment on the content in general terms, but I am not sufficiently familiar with the Python interface to comment on anything specific to that. Take my comments as suggestions.

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
+Distributions:
+
+- :meth:`Graph.degree_distribution`
+- :meth:`Graph.path_length_hist`
+
+Connectedness:
+
+- :meth:`Graph.all_minimal_st_separators`
+- :meth:`Graph.minimum_size_separators`
+- :meth:`Graph.feedback_arc_set`
```
szhorvat, line 206: I would not say that `feedback_arc_set` is a good fit for "Connectedness"

iosonofabio: any other category?

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
+
+- :meth:`Graph.get_adjacency`
+- :meth:`Graph.get_adjacency_sparse` (sparase CSR matrix version)
+- :meth:`Graph.laplacian`
+
+Clustering
+++++++++++
+|igraph| includes several approaches to unsupervised graph clustering and community detection:
+
+- :meth:`Graph.clusters` (aka :meth:`Graph.components`): the connected components
```
szhorvat, line 319: Which one is the primary name here? `clusters` or `components`? I would strongly favour components, as this is the standard graph theory term. "Clusters" is used in other ways too, and [people are already getting confused](https://igraph.discourse.group/t/default-algorithm-for-clustering/348).

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
+To return a directed/undirected copy of the graph, use :meth:`Graph.as_directed` and :meth:`Graph.as_undirected`, respectively.
+
+To permute the order of vertices, you can use :meth:`Graph.permute_vertices`:
+
+>>> g = ig.Tree(6, 2)
+>>> g_perm = g.permute_vertices([1, 0, 2, 3, 4, 5])
+
+The canonical permutation can be obtained via :meth:`Graph.canonical_permutation`, which can then be directly passed to the function above.
+
+To rewire the graph at random while keeping some structural properties, there are:
```
szhorvat, line 348: Mention what properties these two rewiring methods keep invariant (i.e. edge count and degree sequence)?

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
+
+- :meth:`Graph.rewire`
+- :meth:`Graph.rewire_edges`
+
+Line graph
+++++++++++
+To compute the line graph, there is :meth:`Graph.linegraph`:
```
szhorvat, line 355: Say what a "line graph" is.

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
+- :meth:`Graph.cut_vertices` or :meth:`Graph.articulation_points`
+
+Cliques and motifs:
+
+- :meth:`Graph.clique_number` (aka :meth:`Graph.omega`)
+- :meth:`Graph.cliques`
+- :meth:`Graph.maximal_cliques`
+- :meth:`Graph.largest_cliques`
+- :meth:`Graph.motifs_randesu` and :meth:`Graph.motifs_randesu_estimate`
+- :meth:`Graph.g.motifs_randesu_no` counts the number of motifs
```
szhorvat, line 216: Is that `g` there by mistake?

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
+- :meth:`Graph.community_label_propagation`
+- :meth:`Graph.community_leading_eigenvector`
+- :meth:`Graph.community_leading_eigenvector_naive`
+- :meth:`Graph.community_leiden`
+- :meth:`Graph.community_multilevel` (a version of Louvain)
+- :meth:`Graph.community_optimal_modularity` (exact solution, < 100 vertices)
+- :meth:`Graph.community_spinglass`
+- :meth:`Graph.community_walktrap`
+
+Simplification, permutations and rewiring
```
szhorvat, line 333: Is it worth mentioning `is_simple` here once more?

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
+Vice versa, to get the edge if from the source and target vertices, you can use :meth:`Graph.get_eid` or, for multiple pairs of source/targets,
+:meth:`Graph.get_eids`. The boolean version, asking whether two vertices are directly connected, is :meth:`Graph.are_connected`.
+
+To get the edges incident on a vertex, you can use :meth:`Vertex.incident`, :meth:`Vertex.out_edges` and
+:meth:`Vertex.in_edges`. The three are equivalent on undirected graphs but not directed ones, of course:
+
+>>> v = g.vs[0]
+>>> edges = v.incident()
+
+The :meth:`Graph.adjecent` and :meth:`Graph.incident` functions fulfill the same purpose with a slightly different syntax based on vertex ids:
```
szhorvat, line 60: `adjacent`, not `adjecent`. But `adjacent` is deprecated (it is not the correct term for this operation), so it should not be mentioned in any tutorials.

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
+
+Neighborhood
++++++++++++++
+To compute the neighbors, successors, and predecessors, the methods :meth:`Graph.neighbors`, :meth:`Graph.successors` and
+:meth:`Graph.predecessors` are available. The three give the same answer in undirected graphs and have a similar dual syntax:
+
+>>> neis = g.vs[0].neighbors()
+>>> neis = g.neighbors(0)
+
+To get the list of vertives within a certain distance of one or more initial nodes, you can use :meth:`Graph.neighborhood`:
```
szhorvat, line 74: vertives -> vertices

"certain distance *from*" (not of)

Since the sentence started with "vertex", let's not switch midway to "node"

szhorvat: @ntamas On a related note, it would be useful to standardize on a single naming *for the API* (not the documentation). Either use "vertex" consistently (my preference) or use "node" consistently. Do not use both. 

I would not change existing things, but it would be nice to be consistent going forward.

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
+To add nodes to a graph, use :meth:`Graph.add_vertex` and :meth:`Graph.add_vertices`:
+
+>>> g.add_vertex()
+>>> g.add_vertices(5)
+
+This changes the graph `g` in place. You can specify the name of the vertices if you wish.
+
+To remove nodes, use :meth:`Graph.delete_vertices`:
+
+>>> g.delete_vertices(None)  # remove all vertices
```
szhorvat, line 103: Naïve question from someone who doesn't use the Python interface much: How would I figure out from the docs that `g.delete_vertices(None)` deletes *all* vertices?

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
+>>> g_intersection = g and g2
+
+Topological sorting
++++++++++++++++++++
+To sort a graph topologically, use :meth:`Graph.topological_sorting`:
+
+>>> g = ig.Graph.Tree(10, 2, mode=ig.TREE_OUT)
+>>> g.topological_sorting()
+
+Graph traversing
```
szhorvat, line 156: I suggest "traversal" instead of "traversing"

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
++++++++++++++++++++++
+A common operation is traversing the graph. |igraph| currently exposes breath-first search (BFS) via :meth:`Graph.bfs` and :meth:`Graph.bfsiter`:
```
szhorvat, line 158: breath -> breadth

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
+
+>>> [vertices, layers, parents] = g.bfs()
+>>> it = g.bfsiter()  # Lazy version
+
+A depth-first search function is in the works.
```
szhorvat, line 163: Note sure this comment is necessary ...

Can you create a "TODO" issue saying to update this line once DFS has been added? This is the kind of thing that tends to get forgotten. Then we end up with a tutorial that claims that some functionality is NOT available when in reality it is.

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
+- :meth:`Graph.maximum_bipartite_matching` (bipartite graphs)
+
+Other complex measures are:
+
+- :meth:`Graph.assortativity`
+- :meth:`Graph.assortativity_degree`
+- :meth:`Graph.assortativity_nominal`
+- :meth:`Graph.coreness` (aka :meth:`Graph.shell_index`)
+- :meth:`Graph.density`
+- :meth:`Graph.hub_score`
```
szhorvat, line 243: Hub score and authority score should always be mentioned together. See also https://github.com/igraph/igraph/issues/1424

szhorvat: Also, this should go under Vertex properties, not Global properties

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
+- :meth:`Graph.mincut`
+- :meth:`Graph.mincut_value`
+- :meth:`Graph.maximum_bipartite_matching` (bipartite graphs)
+
+Other complex measures are:
+
+- :meth:`Graph.assortativity`
+- :meth:`Graph.assortativity_degree`
+- :meth:`Graph.assortativity_nominal`
+- :meth:`Graph.coreness` (aka :meth:`Graph.shell_index`)
```
szhorvat, line 241: This is a vertex property, not global property.

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
+- :meth:`Graph.density`
+- :meth:`Graph.hub_score`
+- :meth:`Graph.transitivity_undirected`
+- :meth:`Graph.transitivity_local_undirected`
```
szhorvat, line 245: This is a vertex property, not a global property. (Though it makes sense to mention it together with `transitivity_undirected`)

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
+A spectrum of vertex-level properties can be computed. Similarity measures include:
+
+- :meth:`Graph.similarity_dice`
+- :meth:`Graph.similarity_jaccard`
+- :meth:`Graph.similarity_inverse_log_weighted`
+- :meth:`Graph.diversity`
+
+Structural:
+
+- :meth:`Graph.authority_score`
```
szhorvat, line 274: `hub_score` should go here

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
+- :meth:`Graph.betweenness`
+- :meth:`Graph.bibcoupling`
+- :meth:`Graph.closeness`
+- :meth:`Graph.constraint`
+- :meth:`Graph.cocitation`
+- :meth:`Graph.eccentricity`
+- :meth:`Graph.eigenvector_centrality` aka :meth:`Graph.evcent`
```
szhorvat, line 281: @ntamas Is the `evcent` name meant to be used in the future or is it deprecated? Looks like a leftover from a time of less consistent naming ...

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
+- :meth:`Graph.pagerank`
+- :meth:`Graph.personalized_pagerank`
+- :meth:`Graph.strength`
+
+Connectedness:
+
+- :meth:`Graph.subcomponent`
+- :meth:`Graph.is_separator`
+- :meth:`Graph.is_minimal_separator`
+- :meth:`Graph.biconnected_components`
```
szhorvat, line 291: This is not a vertex property (I'd put it under Global)

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
+- :meth:`Graph.blocks`
```
szhorvat, line 292: I can't see `blocks` in the docs

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
+- :meth:`Graph.get_all_simple_paths`
+- :meth:`Graph.spanning_tree` finds a minimum spanning tree
+
+As well as functions related to cuts and paths:
+
+- :meth:`Graph.mincut` calculates the minimum cut between the source and target vertices
+- :meth:`Graph.st_mincut` - as previous one, but returns a simpler data structure
+- :meth:`Graph.mincut_value` - as previous one, but returns only the value
+- :meth:`Graph.all_st_cuts`
+- :meth:`Graph.all_st_mincuts`
```
szhorvat, line 184: Algorithmically, these are related to flows, not to shortest paths. I don't think it's a bad fit here, but I wonder if it makes sense to be a bit redundant and mention things more than once when they are related to multiple areas.

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
+- :meth:`Graph.clique_number` (aka :meth:`Graph.omega`)
+- :meth:`Graph.cliques`
+- :meth:`Graph.maximal_cliques`
+- :meth:`Graph.largest_cliques`
+- :meth:`Graph.motifs_randesu` and :meth:`Graph.motifs_randesu_estimate`
+- :meth:`Graph.g.motifs_randesu_no` counts the number of motifs
+
+Structural:
+
+- :meth:`Graph.edge_connectivity` or :meth:`Graph.edge_disjoint_paths` or :meth:`Graph.adhesion`
```
szhorvat, line 220: This is basically unweighted `mincut_value`. Since "Structural" is so vague, maybe remove this section, and mention these where cuts are discussed? Vertex and edge connectivity are merely the size of the minimum edge and vertex cuts.

doc/source/analysis.rst
szhorvat: This could go together with: 1. topological sorting 2. optimality

szhorvat:
I added a few comments. I hope some are helpful, but it's up to you to make changes or leave things as they are.

doc/source/analysis.rst
szhorvat: @iosonofabio I just saw your question now. Another idea is to make a category for Directed Acyclic Graphs. Things that fit there: `is_dag`, topological sorting, feedback arc set.

doc/source/analysis.rst
szhorvat: I missed the *within*. Using *of* is fine here.

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
 Graph analysis
 ==============
+|igraph| enables analysis of graphs/networks from simple operations such as adding and removing nodes to complex theoretical constructs such as community detection. Read the `API documentation`_ for details on each function and class.
 
-.. note:: TODO. This is a placeholder section; it is not written yet.
+The context for the following examples will be to import the :class:`Graph` class and to have one or more graphs available:
 
+>>> import igraph as ig
```
ntamas, line 7: This is just an anecdotal note: it is very interesting to see how the community has agreed on the _de facto_ abbreviation of igraph just like NumPy became `np` and Pandas became `pd` unanimously. (I never import `igraph` on its own, I always use `from igraph import ...` - but I never use `import numpy as np` or `import pandas as pd` either).

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
+Vertices are numbered 0 to `n-1`, where n is the number of vertices in the graph. These are called the "vertex ids".
+To count vertices, use :meth:`Graph.vcount`:
+
+>>> n = g.vcount()
+
+Edges also have ids from 0 to `m-1` and are counted by :meth:`Graph.ecount`:
+
+>>> m = g.ecount()
+
+To get a sequence of vertices, use their ids and :attr:`Graph.vs`:
```
ntamas, line 30: If we mention IDs, we should illustrate how to use IDs for subsetting `g.vs` (i.e. `g.vs[2]` or `g.vs[2,4,6]`).

doc/source/analysis.rst
ntamas: One other thing that we should agree on is the capitalization of `ids` (or `IDs`). I'm fine with either, I'm not a native English speaker so I can't tell the difference - just let us be consistent.

doc/source/analysis.rst
ntamas: "Vertex" is my preference as well, although another guideline is to use the same name in the Python interface as in the C core. So, for instance, that's why we have `node_compat_fn` and not `vertex_compat_fn`.

doc/source/analysis.rst
ntamas: Hmmm, I did not know that `g.delete_vertices(None)` deletes all the vertices -- it was never intended. It is an unintended consequence of the fact that the function takes something-that-can-be-converted-into-an-igraph_vs_t and that `igraphmodule_PyObject_to_vs_t()` converts `None` to "all vertices".

Maybe we should update `delete_vertices()` to throw an error for `None` as I don't like this syntax. I'd rather have `g.clear()` (but we don't have it now, and it would also need to delete all the _graph_ attributes as well).

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
+Again, you can specify the names or the actual :class:`Vertex` objects instead.
+
+To add edges, use :meth:`Graph.add_edge` and :meth:`Graph.add_edges`:
+
+>>> g.add_edge(0, 2)
+>>> g.add_edges([(0, 2), (1, 3)])
+
+To remove edges, use :meth:`Graph.delete_edges`:
+
+>>> g.delete_edges(None)   # remove all edges
```
ntamas, line 115: Same here - this is an unintended consequence of how we use `igraphmodule_PyObject_to_es_t()` to convert `None` to "all edges". Calling `delete_edges()` with `None` is discouraged and in the future it will probably throw an error.

doc/source/analysis.rst
ntamas: Agreee - let's remove this or convert this into a TODO comment. Most IDEs can be configured to highlight and collect TODO items into a single panel so it won't get forgotten.

doc/source/analysis.rst
ntamas: Yes, that's a mistake.

doc/source/analysis.rst
ntamas: Yes, that's tricky because there's a _global_ transitivity measure and there's also a _local_ (per-vertex) transitivity.

doc/source/analysis.rst
ntamas: I would consider `evcent()` deprecated.

doc/source/analysis.rst
ntamas: `blocks()` is an alias to `biconnected_components()` so they should be mentioned together.

doc/source/analysis.rst
```diff
@@ -1,5 +1,400 @@
...
+and more complex ones:
+
+- :meth:`Graph.edge_betweenness`
+
+Matrix representations
++++++++++++++++++++++++
+Matrix-related functionality includes:
+
+- :meth:`Graph.get_adjacency`
+- :meth:`Graph.get_adjacency_sparse` (sparase CSR matrix version)
```
ntamas, line 312: sparase -> sparse

doc/source/analysis.rst
ntamas: `components()` should be the primary name, but it violates the principle that I tend to follow the naming convention in the C core (where we have `igraph_clusters()`). So I decided to keep both.

doc/source/generation.rst
```diff
@@ -7,6 +7,10 @@ The :class:`Graph` class is the main object used to generate graphs:
 
 >>> from igraph import Graph
 
+To copy a graph, use :meth:`Graph.copy`:
```
ntamas, line 10: Mention that this is a "shallow" copy, i.e. the structure of the graph is copied, but any mutable objects in the attributes are _not_ copied (they would refer to the same instance). Also mention that if someone really needs a deep copy, they should use the `deepcopy` module from Python, which basically does a pickling - unpickling sequence to achieve a deep copy.

src/_igraph/graphobject.c
```diff
@@ -13078,7 +13078,7 @@ struct PyMethodDef igraphmodule_Graph_methods[] = {
    "  the algorithm assumes that there might be some loops in the graph\n"
    "  and calculates the density accordingly. If C{False}, the algorithm\n"
    "  assumes that there can't be any loops.\n"
-   "@return: the reciprocity of the graph."},
+   "@return: the density of the graph."},
```
ntamas, line 13081: Nice catch, thanks!

ntamas:
Added a few comments myself as well.

doc/source/analysis.rst
iosonofabio: Alright created a separate subcategory for DAGs.

doc/source/analysis.rst
iosonofabio: Improved this a little. They don't seem to keep property fixed exactly, just at the distribution level.

doc/source/analysis.rst
iosonofabio: rightio, components is default.

doc/source/analysis.rst
iosonofabio: yep added

doc/source/analysis.rst
iosonofabio: thanks

doc/source/analysis.rst
iosonofabio: yep, added

doc/source/analysis.rst
iosonofabio: yep, removed in favour of `incident`. I use `id` throughout now, since we have functions such as `get_eid`

doc/source/analysis.rst
iosonofabio: Alright, I think `from` reads better. I also switched to `vertices`.

Agreed let's try to prefer `vertex/ices` over `node/s` in the API.

doc/source/analysis.rst
iosonofabio: Haha yeah this one was surprising for me too. I'll remove this from here and start a separate PR implementing `g.clear`.

I don't think the usage for deleting all vertices but leaving the graph attributes will be heavy anyway.

doc/source/analysis.rst
iosonofabio: yep

doc/source/analysis.rst
iosonofabio: indeed

doc/source/analysis.rst
iosonofabio: That PR is done, so I'll just add DFS here and make this PR conditional to #315

doc/source/analysis.rst
iosonofabio: yep good catch

doc/source/analysis.rst
iosonofabio: yep thx

doc/source/analysis.rst
iosonofabio: yep moved

doc/source/analysis.rst
iosonofabio: yep added

doc/source/analysis.rst
iosonofabio: yep deleted

doc/source/analysis.rst
iosonofabio: yep

doc/source/analysis.rst
iosonofabio: moved together as a global prop

doc/source/analysis.rst
iosonofabio: yep added

doc/source/analysis.rst
iosonofabio: yep done!

doc/source/analysis.rst
iosonofabio: hehe I'm like Vincent, like my `ig`

doc/source/analysis.rst
iosonofabio: yep removed, will do a separate PR on that

doc/source/generation.rst
iosonofabio: yeah added

doc/source/analysis.rst
iosonofabio: yes added now

doc/source/analysis.rst
iosonofabio: see also #320

szhorvat: > `igraphmodule_PyObject_to_vs_t()` converts `None` to "all vertices".

Yeah, that reads really funny in English: "Which vertices?" "None" Then it takes *all* of them instead of *none* of them.

ntamas: Okay, the reason is as follows. There are many methods like `degree()` that work on the whole graph by default but that _may_ take a vertex ID or a list of vertex IDs to operate on a subset of vertices instead. For these methods, the `vs` argument needs a default value that we then treat as "all vertices". I took a commonly used shortcut and simply used `None` (which is basically `null` or `nil` in Python), that's why `None` is converted to "all vertices" by `igraphmodule_PyObject_to_vs_t()`. An alternative (and probably a cleaner way) would have been to create a singleton object named `ALL` and then make _this_ the default value instead of `None`. We can add this as a low-priority issue and we can deal with it in the CZI grant if we get to the point where we need to clean up the Python API before 1.0. There would also have to be a deprecation phase where we still allow `None` but show a warning to the user and ask him to use `ALL` instead.

ntamas:
Awesome. Any more comments @szhorvat? Shall we merge this?

szhorvat:
I don't have any more comments.

Issue #295, opened 2020-05-06
DFS: Depth First Search does not exist


michaelasapp:
*Problem*: I do not see that `graph.dfs` is exposed in the Python API to the C++ library.

I see commit b0d74bd54224a3c34febf3e0444f4198a93772d0 that adds BFS and the title include DFS, but the code does not refer to DFS.

rcalfredson:
Coincidentally, just ran into the same problem.

michaelasapp:
I should point out it's certainly not a big deal -- I spent 10 minutes searching the docs, 5 minutes complaining, and 5 minutes writing a small function -- but it was surprising when porting code from R to Python.

FWIW @rcalfredson I just used this.

```python
def dfs(graph, node):
    yield node
    for edge in graph.vs[node].out_edges():
        yield from dfs(graph, edge.target)
```

ntamas:
Sorry guys, will fix this soon (during the weekend if everything goes well).

michaelasapp:
Thank you!

ntamas:
This is taking longer than expected; I was under the impression that `Graph.bfs()` is based directly on `igraph_bfs()` so I could just copy the implementation and replace a few arguments here and there, but it turns out that `Graph.bfs()` is using `igraph_i_bfs()`, an internal and simplified counterpart of `igraph_bfs()` so the code cannot simply be copied.

iosonofabio:
closed via #315

PR #315, opened 2020-07-29
DFS and DFS iterator

iosonofabio:
This PR aims at addressing #295.

In that issue a trivial Python implementation using recursion is proposed. That would be simplest, of course, but potentially runs into a few issues:

- max depth of the Python call stack
- perhaps speed

The way BFS is coded is via a two-tier system:

- a function that returns a whole list of the traversal, which relies on the C core `igraph_i_bfs`
- an iterator that reimplements the basic algorithm (a queue)

Both live at the C API interface with Python and are not recursive, therefore should be more performant than an equivalent, pure Python recursive call.

This PR aims at either of the following three:
1. implementing DFS and its iterator the same way that BFS is implemented
2. take down the whole infrastructure for BFS and substitute it with a simple pure-python recursive call and do the same for DFS
3. leave BFS the way it is and use the pure Python recursion for DFS, along the lines of "better slow than nothing"

Any preferences out there among these?

### DIFF
--- a/src/_igraph/dfsiter.c
+++ b/src/_igraph/dfsiter.c
@@ -1,0 +1,266 @@
+/* -*- mode: C -*-  */
+/* 
+   IGraph library.
+   Copyright (C) 2006-2012  Tamas Nepusz <ntamas@gmail.com>
+   
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 2 of the License, or
+   (at your option) any later version.
+   
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+   
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc.,  51 Franklin Street, Fifth Floor, Boston, MA 
+   02110-1301 USA
+
+*/
+
+#include "dfsiter.h"
+#include "common.h"
+#include "error.h"
+#include "py2compat.h"
+#include "vertexobject.h"
+
+/**
+ * \ingroup python_interface
+ * \defgroup python_interface_dfsiter DFS iterator object
+ */
+
+PyTypeObject igraphmodule_DFSIterType;
+
+/**
+ * \ingroup python_interface_dfsiter
+ * \brief Allocate a new DFS iterator object for a given graph and a given root
+ * \param g the graph object being referenced
+ * \param vid the root vertex index
+ * \param advanced whether the iterator should be advanced (returning distance and parent as well)
+ * \return the allocated PyObject
+ */
+PyObject* igraphmodule_DFSIter_new(igraphmodule_GraphObject *g, PyObject *root, igraph_neimode_t mode, igraph_bool_t advanced) {
+  igraphmodule_DFSIterObject* o;
+  long int no_of_nodes, r;
+
+  o=PyObject_GC_New(igraphmodule_DFSIterObject, &igraphmodule_DFSIterType);
+  Py_INCREF(g);
+  o->gref=g;
+  o->graph=&g->g;
+  
+  if (!PyInt_Check(root) && !PyObject_IsInstance(root, (PyObject*)&igraphmodule_VertexType)) {
+    PyErr_SetString(PyExc_TypeError, "root must be integer or igraph.Vertex");
+    return NULL;
+  }
+  
+  no_of_nodes=igraph_vcount(&g->g);
+  o->visited=(char*)calloc(no_of_nodes, sizeof(char));
+  if (o->visited == 0) {
+    PyErr_SetString(PyExc_MemoryError, "out of memory");
+    return NULL;
+  }
+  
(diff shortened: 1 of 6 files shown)


iosonofabio:
Alright, I implemented two things:

1. `Graph.DFSIter` at the C/Python interface similar to `BFSIter` (basically using a stack instead of a queue)
2. `Graph.dfs` in pure Python but without recursion (also using a stack). This is a temporary implementation until we have a working `igraph_i_dfs` fuction in C, for which I started https://github.com/igraph/igraph/pull/1451. Notice this is not a lazy generator but rather an immediate function: same as `Graph.bfs`. I suppose the `XFSIter` are supposed to be the lazy versions.

I also added tests to for these guys and also one test for `Graph.bfs` which was missing.

There is one thing I am not quite sure about in terms of API. `Graph.bfs` has these `layers`, but it seems like the code was written somewhat confusingly. The C/Python interface file seems to think this vector should have length `n` (number of nodes) and initialized that way, but then the vector gets blissfully cleared inside of `igraph_i_bfs` and has a variable length - the meaning of which escapes me.

Since I didn't understand it I did not add any `layers` to DFS. I don't think it's too important as we can use `DFSIter` to return the distance and parents anyway, but if you have an illuminating explanation I'd love to hear it.

Otherwise ready to merge.

ntamas:
Thanks a lot for this PR! Yes, the problem with BFS and DFS in general is that the "pythonic" way of doing these is via iterators or generator functions, and neither of them can be connected with the C interface because the C core is fundamentally different (it calculates the whole BFS / DFS in advance). That's why we have `.bfs()` and `BFSIter` simultaneously, so the approach you have chosen is correct.

I'll start reviewing this soon.

src/_igraph/dfsiter.c
```diff
@@ -0,0 +1,295 @@
+/* -*- mode: C -*-  */
+/* 
+   IGraph library.
+   Copyright (C) 2006-2012  Tamas Nepusz <ntamas@gmail.com>
```
ntamas, line 4: Please update the copyright note to year 2020, and "The igraph development team". I think that for each source file it is enough to start the copyright from the year that file was added.

src/_igraph/graphobject.c
```diff
@@ -14517,6 +14535,17 @@ struct PyMethodDef igraphmodule_Graph_methods[] = {
...
+   METH_VARARGS | METH_KEYWORDS,
+   "dfsiter(vid, mode=OUT, advanced=False)\n\n"
+   "Constructs a depth first search (DFS) iterator of the graph.\n\n"
+   "@param vid: the root vertex ID\n"
+   "@param mode: either L{IN} or L{OUT} or L{ALL}.\n"
+   "@param advanced: if C{False}, the iterator returns the next\n"
+   "  vertex in DFS order in every step. If C{True}, the iterator\n"
+   "  returns the distance of the vertex from the root and the\n"
+   "  parent of the vertex in the DFS tree as well.\n"
+   "@return: the DFS iterator as an L{igraph.BFSIter} object.\n"},
```
ntamas, line 14548: sed -e 's/BFSIter/DFSIter' :)

src/igraph/__init__.py
```diff
@@ -3211,6 +3211,56 @@ def get_incidence(self, types="type", *args, **kwds):
...
+        stack.append(vid)
+        vids.append(vid)
+        parents.append(vid)
+        added[vid] = True
+
+        # go down the rabbit hole
+        while stack:
+            vid = stack[-1]
+            if mode == IN:
+                edges = self.vs[vid].in_edges()
```
ntamas, line 3244: Can't we simply use `self.neighbors(vid, mode=mode)` here? This would avoid creating a `Vertex` object (the output of `self.vs[vid]`) and would also simplify the code that comes later when we have to figure out which is the "other" endpoint of the edge.

ntamas: Another trick that comes into my mind: we could have another stack where we push the retrieved neighbors of `vid` when it is visited for the first time. Then, whenever we need to find the _next_ neighbor of `vid` to visit, we simply pop off the last element of the list. If the list is empty, there are no more neighbors of `vid` to visit so we can backtrack.

src/_igraph/dfsiter.c
iosonofabio: fixed

src/_igraph/graphobject.c
iosonofabio: hehe touche. fixed

src/igraph/__init__.py
iosonofabio: Alright I use neighbors now, good catch.

I tried playing a little with a different stack with *all* new neighbors, but it didn't make much of a change in terms of speed. I guess we could use a heap to keep the things sorted by distance and parent, but I'm not sure it improves a lot considering it's pure Python anyway.

iosonofabio:
Good, I made the neighbors and all other changes, ready to merge. If you want to elaborate a bit on the "second stack" idea I'm happy to hear, else let's merge.

ntamas:
Okay, let's merge this now and I might experiment a bit with the idea that I outlined above.

ntamas:
FYI, this is what I was thinking about: https://github.com/igraph/python-igraph/commit/5e4a5e2bc10b4ae9d7a9ede6343f55d7e8cdb82a

iosonofabio:
I see, you're caching the function call, looks good

PR #320, opened 2020-08-03
Clear function and improve delete vertices and edges

iosonofabio:
This implements `Graph.clear()` which clears vertices, edges, and attributes.

It also deprecates `Graph.delete_vertices(None)` and `Graph.delete_edges(None)`, opting instead for `Graph.delete_vertices()` and `Graph.delete_edges()`.

Tests still missing, hence draft.

edit: added tests and reviewed, so it should be ready for merge now

### DIFF
--- a/src/_igraph/graphobject.c
+++ b/src/_igraph/graphobject.c
@@ -565,10 +565,21 @@ PyObject *igraphmodule_Graph_add_vertices(igraphmodule_GraphObject * self,
  */
 PyObject *igraphmodule_Graph_delete_vertices(igraphmodule_GraphObject * self,
                                              PyObject * args, PyObject * kwds) {
-  PyObject *list;
+  PyObject *list = 0;
   igraph_vs_t vs;
 
-  if (!PyArg_ParseTuple(args, "O", &list)) return NULL;
+  if (!PyArg_ParseTuple(args, "|O", &list)) return NULL;
+
+  /* no arguments means delete all. */
+  
+  /*Py_None also means all for now, but it is deprecated */
+  if (list == Py_None) {
+        PyErr_Warn(PyExc_DeprecationWarning, "Graph.delete_vertices(None) is "
+                   "deprecated since igraph 0.8.3, please use "
+                   "Graph.delete_vertices() instead");
+  }
+
+  /* this already converts no arguments and Py_None to all vertices */
   if (igraphmodule_PyObject_to_vs_t(list, &vs, &self->g, 0, 0)) return NULL;
 
   if (igraph_delete_vertices(&self->g, vs)) {
--- a/src/_igraph/graphobject.c
+++ b/src/_igraph/graphobject.c
@@ -571,7 +571,7 @@ PyObject *igraphmodule_Graph_delete_vertices(igraphmodule_GraphObject * self,
   if (!PyArg_ParseTuple(args, "|O", &list)) return NULL;
 
   /* no arguments means delete all. */
-  
+
   /*Py_None also means all for now, but it is deprecated */
   if (list == Py_None) {
         PyErr_Warn(PyExc_DeprecationWarning, "Graph.delete_vertices(None) is "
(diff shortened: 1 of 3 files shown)


src/_igraph/graphobject.c
```diff
@@ -565,10 +565,21 @@ PyObject *igraphmodule_Graph_add_vertices(igraphmodule_GraphObject * self,
  */
 PyObject *igraphmodule_Graph_delete_vertices(igraphmodule_GraphObject * self,
                                              PyObject * args, PyObject * kwds) {
-  PyObject *list;
+  PyObject *list = 0;
```
ntamas, line 568: We need to do the same changes in `igraphmodule_Graph_delete_edges()` instead of doing it in the Python implementation of `delete_edges`; the reason is that if we do this in the Python layer, one could still call `GraphBase.delete_edges(graph, None)` to delete all edges.

(I know it's weird, but I have seen novice Python programmers to use the base class like this so we need to cover our bases).

src/igraph/__init__.py
```diff
@@ -382,30 +382,35 @@ def delete_edges(self, *args, **kwds):
         """Deletes some edges from the graph.
 
         The set of edges to be deleted is determined by the positional and
-        keyword arguments. If any keyword argument is present, or the
-        first positional argument is callable, an edge
-        sequence is derived by calling L{EdgeSeq.select} with the same
-        positional and keyword arguments. Edges in the derived edge sequence
-        will be removed. Otherwise the first positional argument is considered
-        as follows:
+        keyword arguments. If the functions is called without any arguments,
```
ntamas, line 385: sed -e s/functions/function/

src/igraph/__init__.py
```diff
@@ -382,30 +382,35 @@ def delete_edges(self, *args, **kwds):
...
+        all edges are deleted. If any keyword argument is present, or the
+        first positional argument is callable, an edge sequence is derived by
+        calling L{EdgeSeq.select} with the same positional and keyword
+        arguments. Edges in the derived edge sequence will be removed.
+        Otherwise the first positional argument is considered as follows:
 
           - C{None} - deletes all edges
+            @deprecated: replaced by no arguments since igraph 0.8.3
```
ntamas, line 393: Not sure if Epydoc likes the `@deprecated` directive in the middle of a list. I would say just spell it out explicitly in a sentence.

src/igraph/__init__.py
```diff
@@ -382,30 +382,35 @@ def delete_edges(self, *args, **kwds):
...
           - a list of 2-tuples - deletes the edges denoted by the given
             source-target vertex pairs. When multiple edges are present
             between a given source-target vertex pair, only one is removed.
         """
         if len(args) == 0 and len(kwds) == 0:
-            raise ValueError("expected at least one argument")
-        if len(kwds)>0 or (hasattr(args[0], "__call__") and \
+            # This means delete all
+            edge_seq = None
+        elif len(kwds) > 0 or (hasattr(args[0], "__call__") and \
```
ntamas, line 403: Now that we are here, can we replace `hasattr(args[0], "__call__")` with `callable(args[0])`? Python did not have `callable()` when I wrote it many years ago, but now it does.

src/_igraph/graphobject.c
iosonofabio: yep, I moved it to the C API now

src/igraph/__init__.py
iosonofabio: yep

src/igraph/__init__.py
iosonofabio: done

src/igraph/__init__.py
iosonofabio: yes

iosonofabio:
I merged current master to force CI. There's one thing with the pandas DataFrame input that I fixed in PR #310 so this PR is now conditional on that one (which is done).

That is, as soon as the refactoring in #310 is done we can merge this one too.

ntamas (approved):

ntamas:
Thanks a lot! PR queue clear, phew!

PR #323, opened 2020-08-10
Clarify Read_EdgeList behavior

remysucre:


szhorvat:
I'll leave all changed to the Python interface (or its docs) to @ntamas, but I would phrase it as:

"Note that the vertex indices are zero-based, therefore the number of vertices in the graph will be the largest integer plus one."

The behaviour is really a consequence of the fact that the integers are vertex *indices* (vertex IDs) and **not** vertex *named*. "Indexing" implies that there is an item for each integer from the smallest (zero) to the largest.

Zero-degree vertices are not really "created". Personally, I find that phrasing misleading. What happens is that if you ever refer to "the 10th vertex" (equivalently, "vertex index 9") that implies that there is also a 1st, 2nd, 3rd, ..., 9th vertex.

ntamas:
I'm okay with the new wording -- technically, @szhorvat is right but this has caused confusion many times in the past (and it still does) so it is probably not pointed out well enough.

PR #329, opened 2020-09-11
GraphSummary: vertex names can be not only str but something else that can be cast to str

deeenes:
measuring explicitly string length when looking for the longest name


ntamas:
Thanks a lot! FYI, a new patch release of igraph and python-igraph is coming soon (hopefully this week).

PR #334, opened 2020-10-01
Adjacency list example to tutorial

kmankinen:
As a someone who recently started to use igraph (instead of networkx), I encountered some missing parts in the documentation. Here's a tiny PR adding to the tutorial.  If you think the viz documentation (visualisation.rst) should also be updated, for example with a basic gallery of graphs using different layouts, I could also help. 

Cheers,
Katja

### DIFF
--- a/doc/source/tutorial.rst
+++ b/doc/source/tutorial.rst
@@ -229,7 +229,7 @@ Let's do the same with a stochastic generator!
 IGRAPH U---- 100 516 --
 + attr: x (v), y (v)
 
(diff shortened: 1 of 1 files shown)


ntamas:
Documentation improvements are always welcome -- thanks a lot! We plan to start improving the documentation in the near future, but in the meanwhile feel free to send PRs about any parts of the documentation where you think it could be improved.

ntamas (approved):

PR #340, opened 2020-11-09
Documentation on graph visualization

iosonofabio:
Hi guys,

As promised I started drafting some docs on graph visualization via `igraph` (well, `python-igraph` I s'pose).

For now it describes a bit the graph layouts, and I am starting to tackle the plotting itself. So about that, it would be great to make a rough plan of action (beyond docs).

First of all:

https://stackoverflow.com/questions/3285193/how-to-change-backends-in-matplotlib-python

Second, I use ipython with matplotlib and `Qt5Agg` backend (it opens up interactive windows, quite convenient). There are a bunch of other interactive backends. How do we use igraph with those? The current docstrings only describe `cairo` or other static backends,but that seems limiting at best given we have the layout already. I usually just use igraph to get a layout and then do the plotting in a little function, e.g.:

```python
import igraph as ig

g = ig.Graph.Full(3)
layout = g.layout('auto')

def plot_graph(g, layout, ax=None):
    x, y = list(zip(*layout.coords))
    if ax is None:
        fig, ax = plt.subplots()
    else:
       fig = None
    ax.scatter(x, y, s=30, color='black', zorder=10)
    for v1, v2 in g.get_edgelist():
        x1, y1, x2, y2 = x[v1], y[v1], x[v2], y[v2]
       ax.plot([x1, x2], [y1, y2], lw=2, color='black', zorder=9)
    if fig is not None:
        fig.tight_layout()
    return {'fig': fig, 'ax': ax}

plot_graph(g, layout)
```

It seems easy enough to do, is it already part of igraph or, if not, should I add it?

Any suggestions on how to write this second part of the doc page? Or should I first implement this kind of plotting function and then finish this doc PR?

Thanks!
Fabio

### DIFF
--- a/doc/source/visualisation.rst
+++ b/doc/source/visualisation.rst
@@ -37,7 +37,7 @@ and methods:
 
 Indexing and slicing can be performed and returns the coordinates of the requested vertices:
 
->>> coords_subgraph = layout[:2]  # Coordinate of the first vertex
+>>> coords_subgraph = layout[:2]  # Coordinates of the first two vertices
 
 .. note:: The returned object is a list of lists with the coordinates, not an `igraph.layout.Layout`
    object. You can wrap the result into such an object easily:
@@ -110,6 +110,13 @@ In both cases, you can add an argument `layout` to the `plot` function to specif
 
 >>> layout = g.layout("kamada_kawai")
 >>> ig.plot(g, layout=layout)
+
+It is also possible to use the name of the layout algorithm directly:
+
+>>> ig.plot(g, layout="kamada_kawai")
+
+If the layout is left unspecified, igraph uses the dedicated `layout_auto()` function, which chooses between one of several
+possible layouts based on the number of vertices and edges.
 
 You can specify additional options such as vertex and edge color, size, and labels via additional arguments, e.g.:
 
--- a/doc/source/visualisation.rst
+++ b/doc/source/visualisation.rst
@@ -2,4 +2,86 @@
 
 Visualisation of graphs
 =======================
-|igraph|
+|igraph| includes functionality to visualize graphs. There are two main components: graph layouts and graph plotting.
+
+In the following examples, we will assume |igraph| is imported as `ig` and a
+:class:`Graph` object has been previously created, e.g.:
+
+>>> import igraph as ig
+>>> g = ig.Graph(edges=[[0, 1], [2, 3]])
+
+Read the `API documentation`_ for details on each function and class.
+
+Graph layouts
++++++++++++++
+A graph *layout* is a low-dimensional (usually: 2 dimensional) representation of a graph. Different layouts for the same
+graph can be computed and typically preserve or highlight distinct properties of the graph itself. Some layouts only make
+sense for specific kinds of graphs, such as trees.
+
+|igraph| offers several graph layouts. The general function to compute a graph layout is :meth:`Graph.layout`:
+
+>>> layout = g.layout(layout='auto')
+
+See below for a list of supported layouts. The resulting object is an instance of `igraph.layout.Layout` and has some
+useful properties:
+
+- :attr:`Layout.coords`: the coordinates of the vertices in the layout (each row is a vertex)
+- :attr:`Layout.dim`: the number of dimensions of the embedding (usually 2)
+
+and methods:
+
+- :meth:`Layout.boundaries` the rectangle with the extreme coordinates of the layout
+- :meth:`Layout.bounding_box` the boundary, but as an `igraph.drawing.utils.BoundingBox` (see below)
+- :meth:`Layout.centroid` the coordinates of the centroid of the graph layout
+
+Indexing and slicing can be performed and returns the coordinates of the requested vertices:
+
+>>> coords_subgraph = layout[:2]  # Coordinate of the first vertex
+
+.. note:: The returned object is a list of lists with the coordinates, not an `igraph.layout.Layout`
+   object. You can wrap the result into such an object easily:
+ 
+   >>> layout_subgraph = ig.Layout(coords=layout[:2])
+
+It is possible to perform linear transformations to the layout:
+
+- :meth:`Layout.translate`
+- :meth:`Layout.center`
+- :meth:`Layout.scale`
+- :meth:`Layout.fit_into`
+- :meth:`Layout.rotate`
+- :meth:`Layout.mirror`
+
+as well as a generic nonlinear transformation via:
+
+- :meth:`Layout.transform`
+
+The following layouts are supported:
+
+- `Graph.layout_bipartite`: bipartite layout
+- `Graph.layout_circle`: circular/spherical layout
+- `Graph.layout_davidson_harel`: Davidson-Harel layout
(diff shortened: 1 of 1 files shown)


ntamas:
Well, the plotting landscape has changed quite a lot since the time we've added plotting functionality to igraph -- right now there are so many options and backends to choose from that I don't really know what the best solution would be.

Back in the old days, Jupyter was in its infancy and so was Matplotlib, so I sort-of assumed that the plotting functionality in igraph has to be standalone because it's unclear who the "winner" will be in the long run (or whether there would be a winner at all). Cairo seemed to be an obvious choice, and the original idea was to support stuff like `plot(g)` in the Python prompt -- this would plot the graph to a temporary file, and open the built-in preferred image viewer of the platform with that temporary file. It sort-of-kind-of worked, apart from a few weird race conditions in certain platforms where the temporary file was removed _before_ the image viewer got a chance to open it. Right now it's still the officially supported way of plotting graphs if you don't have Jupyter.

In Jupyter, the case is a bit different because the `Graph` object has a `_repr_png_()` method that produces a PNG representation of the graph, which can then be shown by Jupyter. (We also have `_repr_svg_()`, but then there's the problem that you have to put each SVG in an `iframe`, otherwise the glyph IDs produced by Cairo would conflict if you have more than one plot in the same Jupyter workbook).

As for your Matplotlib solution, yes, it works as a quick-and-dirty solution, but it does not handle most of the visual attributes that the standard `plot()` method in igraph does. (I'm thinking about stuff like vertex and edge colors, edge weights, labels and so on). So, a long while ago I came up with a [hack that allows igraph graphs in a Matplotlib plot](https://stackoverflow.com/a/36154077/156771). I have no idea whether it still works now or it is outdated. The downside of this solution is that it needs the Cairo backend for Matplotlib, not the Agg backend.

All in all, I think that it would be great to have a solution in `python-igraph` that works nicely with Matplotlib (irrespectively of the chosen plotting backend), but it should also support the visual properties that the standard `plot()` function supports.

ntamas:
As for the second graph: I think the two methods that should definitely be mentioned there are:

1. Simply calling `plot(graph)`, which plots the graph to a temporary file and shows it in the default image viewer of the platform.

2. Calling `plot(graph, "filename.png")` or `plot(graph, "filename.svg")`, which plots the graph to a persistent file, but does not open a window automatically.

iosonofabio:
Thank you. Yes my code was just illustrating the concept. If we want to support matplotlib as a whole, then I'd be happy to make a PR for that that colors things properly.

Personally I think that'd be useful, even though some details (e.g. label positioning in a crowded graph) will be out of scope.

szhorvat:
Just a tip: instead of simply listing layout names, mention in 1-2 words what they are, or what they are good for.

E.g., Sugiyama is for directed acyclic graphs, and for trees

Kamada-Kawai is sometimes described as a "srping layout" since it's based on a physical model of spring

Furchterman-Reingold is a "spring-electric layout", as it's based on a physical model of springs and electrostatic repulsion

graphopt and DrL are also force-directed

Davidson-Harel is based on simulated annealing optimization of a complex function that includes things such as edge crossings

iosonofabio:
Awesome thank you. What is Sugiyama's advantage over other DAG/tree layouts?

szhorvat:
The usual tree layouts work only on trees. Sugiyama works on any DAG. To draw DAGs nicely, one needs to use curved edges. I think this is what it does.

szhorvat:
According to this page, Sugiyama was the first to develop this type of drawing https://en.wikipedia.org/wiki/Layered_graph_drawing

iosonofabio:
Great, I'll add it, thank you!
Fabio

szhorvat:
Some more tips for demonstrating the methods (maybe on a different, more details tutorial page):

For tree layouts, we can use `igraph_tree_game` to generate a random tree. I contributed this function some time ago. I think it's not yet exposed in Python, but it should be.

For DAG layouts (Sugiyama), first we generate an undirected Erdős–Rényi graph, then we convert it to directed with the new `ACYCLIC` method. This is already exposed in Python. This effectively generates a random DAG.

Random trees tend to look pretty good with most of the layouts implemented in igraph: http://szhorvat.net/mathematica/IGDocumentation/#gallery :-)

iosonofabio:
Great ideas! I think that's better in the tutorial, like you said. I'll add a link to the tutorial in here and then we can edit that page separately.

iosonofabio:
@ntamas in a Jupyter notebook, how does one specify to use the PNG representation to avoid the bug? I just opened a notebook and typed:

image

and I got an SVG (I think).

ntamas:
Unfortunately I have no idea; the easiest is to remove the `_repr_svg_` method :) [This page on SO](https://stackoverflow.com/questions/45560556/what-is-the-precedence-order-of-jupyter-rich-representations) claims that all representations are generated by Jupyter (which is a bit unfortunate as we do lots of extra legwork for no reason then), and the most appropriate representation is chosen by the frontend, so in the end you have to look in the frontend configuration to force it to prefer PNG over SVG.

doc/source/visualisation.rst
```diff
@@ -1,4 +1,145 @@
...
+
+and methods:
+
+- :meth:`Layout.boundaries` the rectangle with the extreme coordinates of the layout
+- :meth:`Layout.bounding_box` the boundary, but as an `igraph.drawing.utils.BoundingBox` (see below)
+- :meth:`Layout.centroid` the coordinates of the centroid of the graph layout
+
+Indexing and slicing can be performed and returns the coordinates of the requested vertices:
+
+>>> coords_subgraph = layout[:2]  # Coordinate of the first vertex
```
ntamas, line 40: `layout[:2]` should return the coordinates of the first two vertices; `layout[0]` is the one that returns the coordinates of the first vertex.

doc/source/visualisation.rst
```diff
@@ -1,4 +1,145 @@
...
+
+>>> ig.plot(g, 'myfile.pdf')
+
+.. note:: PNG is a raster image format while PDF, SVG, and Postscript are vector image formats. Choose one of the last three
+   formats if you are planning on refining the image with a vector image editor such as Inkscape or Illustrator.
+
+In both cases, you can add an argument `layout` to the `plot` function to specify a precomputed layout, e.g.:
+
+>>> layout = g.layout("kamada_kawai")
+>>> ig.plot(g, layout=layout)
```
ntamas, line 112: It is also possible to just use the name of the layout algorithm in `ig.plot()`, e.g.: `ig.plot(g, layout="kamada_kawai")`. if you don't add a layout, igraph uses the dedicated `layout_auto()` function, which chooses between one of several possible layouts based on the number of vertices and edges.

doc/source/visualisation.rst
iosonofabio: fixed

doc/source/visualisation.rst
iosonofabio: Yep, I added that info now

iosonofabio:
I added instructions for matplotlib plotting, that makes this PR conditional on #341

ntamas:
Looks good to me! I'll merge this when we are ready with #341

Issue #321, opened 2020-08-04
Plotting: multiple self-loops on a vertex are plotted on top of each other


szhorvat:
When a vertex has multiple self-loops, they are plotted on top of each other. The graphs `0-0` and `0-0, 0-0` are not visually distinguishable when plotted.

These two graphs look the same:

```
plot(Graph([(0,0),(0,0)]))
plot(Graph([(0,0)]))
```

Related: https://github.com/igraph/rigraph/issues/407

szhorvat:
I realize this is not easy. For some inspiration, Mathematica visualizes these like this:

<img width="1202" alt="image" src="https://user-images.githubusercontent.com/1212871/89280874-8b861b80-d649-11ea-994b-9a18c6280a00.png">

iosonofabio:
Is this still present in the matplotlib interface?

ntamas:
The Matplotlib backend seems to be better in handling multiple self-loops; just tested this now in Google Colab.

szhorvat:
> Is this still present in the matplotlib interface?

Yes, here's an example:

```
ig.plot(ig.Graph.Formula('a-b-c-c-c-c-a-b', simplify=False))
```

<img width="512" alt="image" src="https://user-images.githubusercontent.com/1212871/145707577-ee7c7077-942b-4b0a-a78d-ba8a31fe7030.png">

iosonofabio:
Is this the develop branch? If not please test with that

szhorvat:
Yes, it is with the latest develop branch.

szhorvat:
@iosonofabio I remembered correctly, you did fix it. It works in the released version, but it does not work on `develop`. Here's what I get with 0.9.8:

<img width="683" alt="image" src="https://user-images.githubusercontent.com/1212871/145708982-471de2e2-048f-4113-80b6-5ee556d40adb.png">

```
ig.plot(ig.Graph.Formula('a-b-c-c-c-c-a-b', simplify=False), target=plt.axes())
```

PR #341, opened 2020-11-10
Prototype for matplotlib graph drawing

iosonofabio:
Hi guys,

Following up on the chat about plotting documentation, I wrote an initial attempt at enabling graph visualization with matplotlib. That is nice because:

1. a lot of people use matplotlib, so it'd be easy for them to customize the plots if needed
2. several matplotlib backgrounds are interactive, so people can zoom/pan in their graphs

The current plotting framework is designed mostly for static figures rather than interactive ones. Moreover, it relies on hooks in other parts of the code to work, e.g.

- `Graph.__plot__`
- `_repr_svg_` and `_repr_png_` for Jupyter (note: they don't work on vanilla ipython on my machine, but they do work in Jupyter)
- a few magical tricks by Cairo such as autocurving of edges

In order to keep the code somewhat clean, I made a separate class for the matplotlib stuff and implemented a simple plotting routine. It is buggy and lacks some features, but gives you the idea. I'm not strongly attached to the code, but I do think we should help our users make interactive graph visualizations, just opening a static image in an external viewer is quite limiting.

Of course, I acknowledge the amazing work @ntamas and perhaps others have done to get the current version to work: in no way I am trying to diminish your accomplishments! We are just trying to improve on top of what we already have, not to dismantle that.

Now there are a few issues with this idea:

- matplotlib does no magic of bending edges, shrinking edges to fit vertex shape edges, or similia, so one has to decide what to do with that
- matplotlib has a long standing feature request to allow a list of shapes to be passed to the scatter function. For the time being, that makes it difficult (not impossible though) to use more than one shape per graph.
- matplotlib does not have, that I am aware of, a very good way of telling the space occupied on the canvas by text - bases on fontsize, font family, etc. so one has to wing it a little

**NOTE**: matplotlib itself can use Cairo as a drawing engine, for both static or interactive (e.g. 'Qt5Cairo', 'TkCairo') plots. However igraph currently hardlinks to Cairo without the matplotlib layer. That has advantages (less deps, autocurve) but also disadvantages (interactions are impossible, that's what pyplot is essentially for). We could think of a special case when matplotlib is selected in igraph but its backend is actually some form of Cairo: in that case we might be able to get the best of both worlds (e.g. autocurve AND interactivity). Does not seem particularly difficult to implement, but yes a little more laborious than just vanilla pyplot.

What do you think? Feedback welcome, thank you!

### DIFF
--- a/src/igraph/drawing/__init__.py
+++ b/src/igraph/drawing/__init__.py
@@ -27,7 +27,7 @@ import time
 from igraph.compat import property, BytesIO
 from igraph.configuration import Configuration
 from igraph.drawing.colors import Palette, palettes
-from igraph.drawing.graph import DefaultGraphDrawer
+from igraph.drawing.graph import DefaultGraphDrawer, MatplotlibGraphDrawer
 from igraph.drawing.utils import BoundingBox, Point, Rectangle, find_cairo, find_matplotlib
 from igraph.utils import _is_running_in_ipython, named_temporary_file
 
@@ -386,101 +386,6 @@ class Plot(object):
         is drawn"""
         return self.bbox.width
 
-#####################################################################
-
-
-class PlotMatplotlib(object):
-    """Class representing a Matplotlib plot"""
-
-    _shape_dict = {
-        'rectangle': 's',
-        'circle': 'o',
-        'hidden': 'none',
-        'triangle-up': '^',
-        'triangle-down': 'v',
-    }
-
-    def __init__(self, ax=None):
-        """Creates a new plot.
-
-        @param ax: the matplotlib.pyplot.Axes instance to plot the graph into.
-          If this argument is None, a new figure and axes will be created.
-        """
-        if ax is None:
-            fig, ax = plt.subplots()
-
-        self.ax = ax
-
-    def add(self, obj, *args, **kwds):
-        self._objects.append((obj, args, kwds))
-
-    def show(self):
-        import matplotlib.markers as mmarkers
-
-        ax = self.ax
-
-        for obj, args, kwds in self._objects:
-            # FIXME: deal with unnamed *args
-
-            # Get layout
-            layout = kwds.get('layout', obj.layout())
-            if isinstance(layout, str):
-                layout = obj.layout(layout)
-
-            # Vertex coordinates
-            vcoord = layout.coords
-
-            # Edge coordinates
-            ecoord = []
-            for edge in obj.es:
-                src, tgt = edge.source, edge.target
-                src_coord = vcoord[src]
-                tgt_coord = vcoord[tgt]
-                ecoord.append((src_coord, tgt_coord))
-
-            # Vertex properties
-            # NOTE: shapes use slightly different names from Cairo
-            s = kwds.get('vertex_size', None)
-            c = kwds.get('vertex_color', None)
-            label = kwds.get('vertex_label', None)
-            label_size = kwds.get('vertex_label_size', None)
-            zorder = kwds.get('vertex_order')
-            shapes = kwds.get('vertex_shape', None)
-            if shapes is not None:
-                if isinstance(shapes, str):
-                    shapes = self._shape_dict.get(shapes, shapes)
-                elif isinstance(shapes, mmarkers.MarkerStyle):
-                    pass
-
-            # Scatter vertices
-            # NOTE: matplotlib does not support a list of shapes yet
-            x, y = list(zip(*vcoord))
-            ax.scatter(x, y, s=s, c=c, m=shapes, zorder=zorder)
-            if label is not None:
-                for i, lab, lab_size in enumerate(zip(label, label_size)):
-                    xi, yi = x[i], y[i]
-                    ax.text(xi, yi, lab, fs=lab_size)
-
-            # Edge properties
-            edge_width = kwds.get('edge_width', None)
-            arrow_width = kwds.get('arrow_width', None)
-            arrow_length = kwds.get('arrow_size', None)
-            ec = kwds.get('edge_color', None)
-
-            # Plot edges
-            for ((xs, ys), (xt, yt)) in ecoord:
-                if obj.is_directed():
-                    dx = xt - xs
-                    dy = yt - ys
-                    ax.arrow(
-                        xs, ys, dx, dy, length_includes_head=True,
-                        width=edge_width, head_width=arrow_width,
-                        head_length=arrow_length,
-                        color=ec,
-                    )
-                else:
-                    ax.plot([xs, xt], [ys, yt], lw=edge_width, c=ec)
-
 
 #####################################################################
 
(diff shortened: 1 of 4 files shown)


ntamas:
Sorry for the delay; finally I managed to take a look at the first draft and I think it's very good as a starting point!

One thing that I would try at this point is to see whether this could be reformulated as a subclass of `AbstractGraphDrawer`, which is not much more than an interface specification for objects that draw graphs _somehow_. `DefaultGraphDrawer`, the current graph drawer that igraph uses is a subclass of `AbstractGraphDrawer`, so it would probably make sense to implement the Matplotlib-based graph drawer as a subclass of `AbstractGraphDrawer` as well.

The next step would probably be to make sure that most of the "standard" vertex / edge attributes that the `DefaultGraphDrawer` class supports for styling vertices / edges is also supported by the Matplotlib graph drawer. I don't know much about the capabilities of Matplotlib so I have no idea how difficult this can be; if it is too much of work at this stage, I'm okay with having a simplistic Matplotlib graph drawer in igraph first and then move on to handling the more complex cases later.

iosonofabio:
After a busy time, I had a deeper look at the interface today. I realized a lot of things that I thought were done in Cairo (e.g. autocurve) are actually coded out in igraph, @ntamas that must have been a nightmare back in the day! Big kudos for that!

So the good news is that we can in principle recycle that for the matplotlib interface. The other good news is as you say there are the abstract classes so one can subclass for matplotlib.

The bad news is that even abstract classes for vertex, edge, graph, text and that uses cairo in a million places. E.g. `AbstractEdgeDrawer` takes a `context` in the constructor, which basically means cairo context and does not have a direct equivalent in matplotlib (as far as a simple interface goes, at least). The other thing is that generic functions and classes (e.g. for calculating the bezier cubic) are sometimes nested within concrete cairo class methods, making them hard to access. Finally, diamond inheritance is commonly used, which is typically a little messy to adapt.

So I will try to move along the following lines:

1. keep it simple: make a somewhat complete and working interface for matplotlib without trying to emulate every single feature of the cairo interface and without making too many new classes.
2. functionality over elegance: copy paste some code (e.g. bezier cubic) even though it leads to code duplication
3. respect the work: I won't start modifying the gears of the current implementation with its metaclasses, diamond inheritance, etc. @vtraag mentioned we might aim for a graph drawing library beyond igraph anyway, perhaps in collaboration with other graph tools, so streamlining our current implementation might not be so useful in that perspective.

I'll let you know when this needs another round of review, thank you!

iosonofabio:
Alright, it now works on my machine with simple option, e.g.:

```python
import igraph

g = igraph.Graph.Famous('zachary')
g.to_directed()

fig, ax = plt.subplots(figsize=(3, 3))
igraph.plot(
    g,
    target=ax,
    vertex_size=50, edge_order=3, vertex_order=2, edge_curved=0.5, arrow_width=2, arrow_length=4,
    edge_color='k', vertex_color='steelblue', edge_alpha=0.3,
)

plt.ion()
plt.show()
```

As far as I can tell, the main things missing are:
- changing a few options on a verex or edge-basis (e.g. some vertices are squares, some are circles) - this is directly coupled to matplotlib's own limitations in that respect
- edges always reach the center of the vertex, without shrinking until the border of the *shape* representing the vertex. The cairo version has an in-house binary search for this. I'm a little afraid this is not easily comparible with matplotlib's interactive features, because things like marker size autoscale when zoomed into
- vertex label positioning could be improved

szhorvat:
Just wanted to point you to #321 in case this PR leads to changes to the design that will make it more difficult / easier to fix that bug (I have no idea, didn't follow the details).

iosonofabio:
> Just wanted to point you to #321 in case this PR leads to changes to the design that will make it more difficult / easier to fix that bug (I have no idea, didn't follow the details).

Excellent point @szhorvat 

The self loop is still missing from mine. The function for the cairo plotting is pretty straightforward. The problem is that Mathematica changes the layout based on self loop, whereas here we typically come in with a layout already. But let me see

szhorvat:
Mathematica doesn't change the vertex positions ("layout") based on the loops, but it does change both the loop directions and sizes (narrow/wide) based on the vertex positions (it adapts them to the direction of the adjacent non-loop edges). 

It would be great if the same could be achieved with Python, but as a start it would be already a great improvement if multiple loops didn't overlap. I.e., compute the loop directions/sizes simply based on how many loops there are in total on the vertex.

I have no idea if this is easy to do at the moment, and I definitely don't want to hijack the matplotlib backend PR. I just mentioned this in case keeping it in mind would change how you design the matplotlib backend.

iosonofabio:
Thank you for the clarification. I see that Mathematica puts all loops between two consecutive (in terms of angles) non-loops. Basically you identify the largest pizza slice, divide it by (nloops - 1), and build a bunch of cubic Béziers. Shouldn't be horribly difficult, will give it a go

iosonofabio:
I implemented a simple if imperfect version of the loops thing @szhorvat has asked for:

plot_mpl_withloops

A few things still outstanding (e.g. finite size of vertices) but I think we could merge already after I add a few tests. Assuming we need tests for this kind of stuff?

szhorvat:
That looks pretty great!

Does it handle multiple self-loops on the same vertex too?

szhorvat:
I don't know matplotlib very well, but I was wondering if it is possible to not distort those self-loops with the aspect ratio. I see that the vertices are perfectly round, but the loops look a bit stretched.

There's also offsetting edges so that the arrowheads would touch the perimeter of circles instead of their centre.

Of course it's easy for me to point out small flaws, not so easy to fix them :-)

iosonofabio:
Yes it handles multiple loops:

plot_mpl_withloops2

They are already stretched by the aspect ratio. The main thing I should figure out is where to put the Bezier waypoints so that they look pretty all the time: not super interesting as you can imagine.

The thing of offsetting by the dot size is pretty easy EXCEPT for zoom-in interactions:

plot_mpl_withloops2_zoomin

As you can see the size of the vertex dot is unchanged, i.e. matplotlib does *autoscaling* on scatter plots. If we just use data coordinates to offset the arrows, the offset will not easily autoscale unless we code a separate function that is triggered by the zoom-in event and recalculated the offsets for all vertices based on the new (in data coordinates, smaller) vertex dot size.

Not that writing the function per se is hard, but one has to be careful to not replot any arrow but only atomically touch the properties (i.e. start and end (x,y) coordinates). Otherwise, if someone plots, changes color of the arrows, then zooms in, the color is reset to the original color.

I'll give it a go, but am not sure how far it will go.

iosonofabio:
Alright, I think this fixes it to the extent that it matters. I've added:

- a correction for finite-size vertices that autoscales with zooms
- better heuristics for cubic Bezier lines for loops

plot_mpl_withloops3
plot_mpl_withloops3_zoomin

I think it's good enough to be useful for a number of people already. One thing I could add would be to return something useful instead of None, e.g. a dictionary with the labels, vertex, and edge handles for further customization.

ntamas:
Thanks @iosonofabio , very nice! I'll review it carefully soon (tomorrow at the latest).

src/igraph/drawing/__init__.py
```diff
@@ -482,7 +490,7 @@ def plot(obj, target=None, bbox=(0, 0, 600, 600), *args, **kwds):
     # so just show or save the result
     if target is None:
         result.show()
-    elif isinstance(target, basestring):
```
ntamas, line 485: This breaks compatibility with Python 2.7 where filenames could be instances of `bytes` or `str`, but I think we will drop support for 2.7 anyway soon so I don't mind. The same applies to occurrences of `xrange`.

ntamas (approved):

ntamas:
Thanks a lot!

iosonofabio:
Awesome thanks! The breaking changes were on purpose 😉

Issue #317, opened 2020-07-29
Overhaul installation instructions on website


szhorvat:
Some concerns about the installation instructions on the website:

https://igraph.org/python/#startpy

 - Should it mention conda? I was surprised that it does not, since @vtraag is actively supporting it. https://anaconda.org/conda-forge/python-igraph
 - It says, use 0.8.0 with Python 2.7, but as far as I understand, we didn't drop 2.7 support in the 0.8 series. So it should day `0.8` and not `0.8.0`.

iosonofabio:
We could mention conda, why not. And as of the other change, definitive EOL for Pyhton 2.7 is end of 2020, so yes we can change it into 0.8 or change it to 0.8.3 (the next minor release) since that will most likely be the last one we get through.

Oh I can't wait letting that stone fall.

ntamas:
I have zero experience with Conda so that's why it was not included so far. If @vtraag is committed to supporting it for the foreseeable future, I am happy to include a short paragraph about it - just send a PR if there are any Conda-specific instructions that I should add.

As for the 0.8.0 vs 0.8 issue, I have now replaced 0.8.0 with 0.8.x so it should be okay.

vtraag:
I will add a paragraph on `conda` in the near future. Good idea to include it in the installation instructions indeed.

iosonofabio:
Hi guys,

It was recently brought to my attention by the networkx/networkx#4404 folks that our "Installation from source" instructions are currently somewhat terse:

https://igraph.org/python/doc/tutorial/install.html#compiling-igraph-from-source

Since there's ont PR from @vtraag ATM I'll make one that at least includes some info on that and streamlines the rest a little. If that's not ok please let me know.

ntamas:
See my comments on #351; in theory, the installation should be much simpler (no need to install `igraph` separately), and if it does not work, it's a bug and we need to work on that -- the installation instructions should still say that `pip install` should work.

PR #350, opened 2020-12-06
Install from source instructions

iosonofabio:
Trying to tackle #317, better installation instructions. For now I worked on the "Install from source" section.


iosonofabio:
Ooops started from the wrong repo, closing

PR #351, opened 2020-12-06
Install from source instructions

iosonofabio:
Trying to tackle #317, better installation instructions. For now I worked on the "Install from source" section.

Now PR comes from the right repo, so as not to pollute the branch namespace.

### DIFF
--- a/doc/source/install.rst
+++ b/doc/source/install.rst
@@ -137,7 +137,52 @@ Package Index <http://pypi.python.org/pypi/python-igraph>`_ instead to get a mor
 release (see `Installing igraph from the Python Package Index`_).
 
 
-Compiling |igraph| from source
-==============================
+Compiling |python-igraph| from source
+========================================
 
-TODO
+|python-igraph| has C Python binding into the main |igraph| library, so you'll need both a C compiler
+and the library itself. If you do not have the C |igraph| library on your system, the installer
+will try to download and compile it on the fly, however dynamic linking against an existing library
+is preferred if available. Generally speaking, it's best not to compile a recent |python-igraph|
+version against a much older |igraph| C library as both are under active development.
+
+|igraph| library
+----------------
+To install the |igraph| C library, you would generally follow the official `documentation <https://igraph.org/c/>`,
+however on many linux distributions a precompiled |igraph| binary is available through your
+package manager, e.g. on Arch linux:
+
+  $ pacman -S igraph
+
+(you need root privileges to install packages on most systems).
+
+Download |python-igraph| source code
+------------------------------------
+If you are looking for the latest release, you can find it on `Pypi <https://pypi.org/project/python-igraph/#files>`
+or `Github <https://github.com/igraph/python-igraph/releases/>` and decompress it. If you are looking for the latest git
+master branch, clone the repo:
+
+  $ git clone https://github.com/igraph/python-igraph.git
+
+Then go into the folder:
+
+  $ cd python-igraph
+
+(it might have a slightly different name depending on the release).
+
+Installing using pip
+--------------------
+Run from inside the folder:
+
+  $ pip install .
+
+Compiling using setup.py
+------------------------
+Run from inside the folder:
+
+  $ python setup.py build
+
+The compiled library will be in the folder `build/lib.<your system>-<your Python version>`. You can add that folder to
+your `PYTHONPATH` or, if you prefer to install the library properly:
+
+  $ python setup.py install
--- a/doc/source/install.rst
+++ b/doc/source/install.rst
@@ -40,7 +40,7 @@ and macOS. Currently there are binary packages for Python 2.7, and Python 3.5 th
 note that support for Python 2.7 will be discontinued with the version 0.9.0 release of
 |igraph|'s Python interface.
 
-To install |python-igraph| globally, use the followig command (you probably need
+To install |python-igraph| globally, use the following command (you probably need
 administrator/root priviledges):
 
   $ pip install python-igraph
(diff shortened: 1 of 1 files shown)


ntamas:
We need to deal with two cases of "compiling from source"; the first is to "compile the most recent development version from Github to get all the bleeding edge stuff", and the second is to "compile the source tarball from PyPI on obscure platforms where we don't have a Python wheel yet". I'll start with the first.

### Compiling the dev version from Github

The `python-igraph` repo always contains a submodule reference to the matching `igraph` version that it is supposed to compile with, so there is no need to install `igraph` separately -- in fact, if you are compiling `python-igraph` from Github, the best is _not_ to have `igraph` installed in advance to avoid cases when the compilation process somehow "picks up" the pre-installed `igraph` version and not the one that is included in the `python-igraph` repo as a submodule.

What you were missing is that after running `git clone` to get a recent copy of `python-igraph`, you need to run `git submodule update --init` to ask Git to check out all submodules as well. Once that is done, `python setup.py install` will (should) use the vendored submodule.

Furthermore, if you run `pip install git+https://github.com/igraph/python-igraph`, `pip` is smart enough to initialize the submodules automatically. Therefore, I would simply recommend `pip install git+https://github.com/igraph/python-igraph`. If that doesn't work for some reason, that's a bug, but in that case you can still try `git clone` followed by `git submodule update --init`.

### Compiling a source tarball on obscure platforms

The source tarballs published on PyPI come with a vendored copy of the matching `igraph` version, so again there is no need to install igraph separately -- simply run `pip install python-igraph`, and it will go to PyPI, fetch the source (since it won't find a matching pre-compiled wheel), extract it and then attempt to compile it. This should also work out of the box. Again, there is no need to install the C core separately, the matching C core is vendored into the tarball.

iosonofabio:
Thank you Tamas. I'll integrate that into the PR.

Some people (like me) prefer to install the igraph core using their package manager and then compile python-igraph on top of that. The installer currently tells me no detected igraph library was found, but then I press enter and it happily compiles and links against it? That is probably a bug - a good bug, for once ;-)

The other thing is that although pip is great, I like the idea of telling people about the manual procedure with submodule and setup.py because it removes a layer of magic. So I'll try to include both explanations.

Thank you!

vtraag:
> install the igraph core using their package manager

Note that you may not get the most recent `igraph` C core, depending on how quickly the package manager included a new release. Perhaps `pacman` is really fast with including new releases, I'm not sure. As long as it is at least a 0.8.x release all versions should be compatible. I do believe you need the flag `--pkg-config` to enforce the python interface to link to the external library, you might want to double check that.

iosonofabio:
That's a common discussion point. Arch linux prides itself to being quite fast in these things, but of course that's often not the case for many distros. I'll try to point that out.

As of the second point, I think there's currently a bug. I do not use `--pkg-config` and indeed the installer prints a message to screen saying that it looked into two folders for `igraph` headers and didn't find anything, however it must have because it then compiled against those very headers if I just press Enter...

szhorvat:
> Some people (like me) prefer to install the igraph core using their package manager and then compile python-igraph on top of that.

I'm still of the opinion that this should not be encouraged. I understand that some Linux distros are dead set on "zero duplication", so the possibility should be supported (so that they can have their python-igraph link to their igraph), but it should not be the default.

Tying the releases of different interfaces together is going to be an inconvenient constraint with no tangible benefit. It should be possible to make a release of the C core without having to wait for python-igraph to be ready. The various package managers (MacPorts, Homebrew, etc.) should be able to upgrade their C igraph without having to wait for python-igraph to also make a release.  python-igraph should be able to make a new release fixing Python-only stuff without the inconvenient constraint of having to match version numbers with the C core.  We should not waste effort on trying to support a python-igraph linked to an _arbitrary_ C core, and try to figure out why a user is seeing problem when everything works on our computers: "Use a _matching_, officially supported C core, or you're on your own."

Note that neither the R interface nor the Mathematica one is fully in sync with C core releases. Furthermore, they both link to the C core statically. This is true for R/igraph even when packages by Linux distros—it does not use the distro's igraph package (and attempting to make it use it would create a ton of extra work for us without providing any benefits, just think all that `USING_R` and the need for CRAN-conformance).

vtraag:
I agree completely with @szhorvat. However, we should perhaps enable a simply major version check in interfaces (e.g. interface is compatible with release series 0.8.x).

> think all that USING_R and the need for CRAN-conformance

This is not the case for system libraries, CRAN conformance is not enforced for system libraries.

ntamas:
> I'm still of the opinion that this should not be encouraged. 

I also agree with that; the only way I can guarantee that `python-igraph` compiles is that I pinpoint a specific version of the C core library that it works with, and that's exactly what the Git submodule does, for every revision. I try to align these with "real" igraph releases whenever I make a release of `python-igraph`, though.

> I understand that some Linux distros are dead set on "zero duplication", so the possibility should be supported (so that they can have their python-igraph link to their igraph), but it should not be the default.

I think @iosonofabio was not advocating for this to be the default, he would just like to include more details about the process in the documentation so if someone really insists on linking `python-igraph` to a specific version of `igraph`, he gets an overview of how this could be done. I'm totally fine with that -- I would just like to make it loud and clear that there are no guarantees that the Python interface will work when doing so. Officially, only that specific revision of the C core is supported that corresponds to a particular commit of the Python interface.

vtraag:
> Officially, only that specific revision of the C core is supported that corresponds to a particular commit of the Python interface.

I think that if we take out versioning approach seriously, `python-igraph` should compile with every release from a specific major release series, right? That is, in this case, it should compile with all versions of 0.8.x in this case.

ntamas (approved):

ntamas:
Yes @vtraag , in theory it should, but at this point I would not like to make claims about that in the documentation :)

iosonofabio:
Agreed. I would just like to tell people what practices are used by folks, not encourage them to try and be weird.

Happy to rephrase adding notes like "This is not officially supported, YMMV" inthat section.

Issue #353, opened 2021-01-01
Adding an image viewer for FreeBSD


thierry-FreeBSD:
python-igraph works fine on FreeBSD, but the configuration for an image viewer is missing.
This simple patch provides it:
```
--- src/igraph/configuration.py.orig	2020-10-08 10:13:28 UTC
+++ src/igraph/configuration.py
@@ -56,6 +56,18 @@ def get_platform_image_viewer():
                 if os.path.isfile(full_path):
                     return full_path
         return ""
+    elif plat == "FreeBSD":
+        # FreeBSD also has a whole lot of choices, try to find one
+        choices = ["eog", "gthumb", "geeqie", "display",
+                   "gpicview", "gwenview", "qiv", "gimv", "ristretto",
+                   "geeqie", "eom"]
+        paths = ["/usr/local/bin"]
+        for path in paths:
+            for choice in choices:
+                full_path = os.path.join(path, choice)
+                if os.path.isfile(full_path):
+                    return full_path
+        return ""
     elif plat == "Windows" or plat == "Microsoft":    # Thanks to Dale Hunscher
         # Use the built-in Windows image viewer, if available
         return "start"
```

ntamas:
Awesome, thanks a lot!

Can you submit this as a PR so you can get proper credit for it in the commit log automatically when I merge it? Alternatively, I can add it myself, but then I can only mention your name in a comment.

PR #354, opened 2021-01-02
Adding an image viewer for FreeBSD

thierry-FreeBSD:
See https://github.com/igraph/python-igraph/issues/353


ntamas:
Thanks a lot!

PR #355, opened 2021-01-10
Get pandas DataFrame for vertices and edges

iosonofabio:
Adding two methods to `Graph` to export vertices/edges and their names and attributes to pandas DataFrames. Basically vincent code from here:

https://igraph.discourse.group/t/convert-graph-object-to-dataframe/575

with a few changes for corner cases and tests.

Somehow my tests on 2.7 pass but on 3.9 complain they cannot import pandas. I have pandas in my python 3.9 installation, so... well let's see how CI goes.

### DIFF
--- a/src/igraph/__init__.py
+++ b/src/igraph/__init__.py
@@ -53,7 +53,6 @@ import sys
 import operator
 
 from collections import defaultdict
-
 from shutil import copyfileobj
 from warnings import warn
 
@@ -3409,6 +3408,76 @@ class Graph(GraphBase):
                     e[a_name] = a_value
 
         return g
+
+    def get_vertex_dataframe(self):
+        """Export vertices with attributes to pandas.DataFrame
+
+        @return: a pandas.DataFrame representing vertices and their attributes.
+          The index uses vertex IDs, from 0 to N - 1 where N is the number of
+          vertices.
+
+        If you want to use vertex names as index, you can do:
+
+        >>> df = graph.get_vertex_dataframe()
+        >>> df.set_index('name', inplace=True)
+
+        """
+        try:
+            import pandas as pd
+        except ImportError:
+            raise ImportError("You should install pandas in order to use this function")
+
+        df = pd.DataFrame(
+            {attr: self.vs[attr] for attr in self.vertex_attributes()},
+            index=list(range(self.vcount())),
+        )
+        df.index.name = "vertex ID"
+
+        return df
+
+    def get_edge_dataframe(self):
+        """Export edges with attributes to pandas.DataFrame
+
+        @return: a pandas.DataFrame representing edges and their attributes.
+          The index uses edge IDs, from 0 to M - 1 where M is the number of
+          edges. The first two columns of the dataframe represent the IDs of
+          source and target vertices for each edge. These columns have names
+          "source" and "target". If your edges have attributes with the same
+          names, they will be present in the dataframe, but not in the first
+          two columns.
+
+        If you want to use source and target vertex IDs as index, you can do:
+
+        >>> df = graph.get_edge_dataframe()
+        >>> df.set_index(['source', 'target'], inplace=True)
+
+        The index will be a pandas.MultiIndex. You can use the `drop=False`
+        option to keep the `source` and `target` columns.
+
+        If you want to use vertex names in the source and target columns:
+
+        >>> df = graph.get_edge_dataframe()
+        >>> df_vert = graph.get_vertex_dataframe()
+        >>> df['source'].replace(df_vert['name'], inplace=True)
+        >>> df['target'].replace(df_vert['name'], inplace=True)
+        >>> df_vert.set_index('name', inplace=True)  # Optional
+
+        """
+        try:
+            import pandas as pd
+        except ImportError:
+            raise ImportError("You should install pandas in order to use this function")
+
+        df = pd.DataFrame(
+            {attr: self.es[attr] for attr in self.edge_attributes()},
+            index=list(range(self.ecount())),
+        )
+        df.index.name = "edge ID"
+
+        df.insert(0, "source", [e.source for e in self.es], allow_duplicates=True)
+        df.insert(1, "target", [e.target for e in self.es], allow_duplicates=True)
+
+        return df
 
     def bipartite_projection(
         self, types="type", multiplicity=True, probe1=-1, which="both"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -38,6 +38,9 @@ install:
   # install cibuildwheel
   - pip install cibuildwheel==1.6.1
 
+  # install pandas (optional)
+  - pip install pandas
+
 before_build:
   - git submodule update --init --recursive
 
--- a/tests/test_foreign.py
+++ b/tests/test_foreign.py
@@ -4,7 +4,12 @@ import warnings
 
 from igraph import *
 
-from .utils import temporary_file
+from .utils import temporary_file, skipIf
+
+try:
+    import pandas as pd
+except ImportError:
+    pd = None
 
 
 class ForeignTests(unittest.TestCase):
--- a/tests/test_foreign.py
+++ b/tests/test_foreign.py
@@ -335,6 +335,84 @@ class ForeignTests(unittest.TestCase):
             self.assertTrue(g.vcount() == 3 and g.ecount() == 1 and not g.is_directed())
             g.write_pickle(tmpfname)
 
+    @skipIf(pd is None, "test case depends on Pandas")
+    def testVertexDataFrames(self):
+        g = Graph([(0, 1), (0, 2), (0, 3), (1, 2), (2, 4)])
+
+        # No vertex names, no attributes
+        df = g.get_vertex_dataframe()
+        self.assertEqual(df.shape, (5, 0))
+        self.assertEqual(list(df.index), [0, 1, 2, 3, 4])
+
+        # Vertex names, no attributes
+        g.vs["name"] = ["eggs", "spam", "ham", "bacon", "yello"]
+        df = g.get_vertex_dataframe()
+        self.assertEqual(df.shape, (5, 1))
+        self.assertEqual(list(df.index), [0, 1, 2, 3, 4])
+        self.assertEqual(list(df["name"]), g.vs["name"])
+        self.assertEqual(list(df.columns), ["name"])
+
+        # Vertex names and attributes (common case)
+        g.vs["weight"] = [0, 5, 1, 4, 42]
+        df = g.get_vertex_dataframe()
+        self.assertEqual(df.shape, (5, 2))
+        self.assertEqual(list(df.index), [0, 1, 2, 3, 4])
+        self.assertEqual(list(df["name"]), g.vs["name"])
+        self.assertEqual(set(df.columns), set(["name", "weight"]))
+        self.assertEqual(list(df["weight"]), g.vs["weight"])
+
+        # No vertex names, with attributes (common case)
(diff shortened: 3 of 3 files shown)


ntamas (approved):

ntamas:
Looks good to me! Is this ready for merge?

vtraag:
I have a few comments, adding my review now.

src/igraph/__init__.py
```diff
@@ -3190,6 +3190,54 @@ def DataFrame(klass, edges, directed=True, vertices=None, use_vids=False):
...
+          the dataframe, but not in the first two columns.
+        """
+        import pandas as pd
+
+        df = pd.DataFrame(
+            {attr: self.es[attr] for attr in self.edge_attributes()},
+            index=list(range(self.ecount())),
+        )
+        if 'name' in df.columns:
+            df.set_index('name', inplace=True)
```
vtraag, line 3232: I don't think we ever use the `name` attribute for `edges` anywhere else, right? I think it might be strange to start using it here.

Instead, I would suggest that we add the `source` and `target` columns as index.

src/igraph/__init__.py
```diff
@@ -3190,6 +3190,54 @@ def DataFrame(klass, edges, directed=True, vertices=None, use_vids=False):
...
+
+        df.insert(
+            0, 'source', [e.source for e in self.es], allow_duplicates=True)
```
vtraag, line 3235: It might make sense here to use the `name` of the `node_df`, because that is effectively also the `index` of the `vertex_df`.

src/igraph/__init__.py
```diff
@@ -3190,6 +3190,54 @@ def DataFrame(klass, edges, directed=True, vertices=None, use_vids=False):
...
+        df.insert(
```
vtraag, line 3236: Likewise here of course.

src/igraph/__init__.py
```diff
@@ -3190,6 +3190,54 @@ def DataFrame(klass, edges, directed=True, vertices=None, use_vids=False):
 
         return g
 
+    def get_vertex_dataframe(self):
```
vtraag, line 3193: Perhaps we could make it optional to use the `name` attribute as the index or not? There are a few options here. We could allow the user to specify an attribute as index, or simply a boolean indicating or not whether the vertex `name` attribute should be used for creating the dataframe.

src/igraph/__init__.py
```diff
@@ -3190,6 +3190,54 @@ def DataFrame(klass, edges, directed=True, vertices=None, use_vids=False):
...
+        """Export vertices with attributes to pandas.DataFrame
+
+        @return: a pandas.DataFrame representing vertices and their attributes.
+          If the vertices have names, the index of the dataframe contains them,
+          otherwise it's integers from 0 up (which are also the vertex IDs).
+          The order of the vertices is the same as in the graph, so the n-th
```
vtraag, line 3199: Typo:
```suggestion
          The order of the vertices is the same as in the graph, so the i-th
```

src/igraph/__init__.py
```diff
@@ -3190,6 +3190,54 @@ def DataFrame(klass, edges, directed=True, vertices=None, use_vids=False):
...
+            df.set_index('name', inplace=True)
+        return df
+
+    def get_edge_dataframe(self):
+        """Export edges with attributes to pandas.DataFrame
+
+        @return: a pandas.DataFrame representing edges and their attributes.
+          If the edges have names, the index of the dataframe contains them,
+          otherwise it's integers from 0 up (which are also the edge IDs).
+          The order of the edges is the same as in the graph, so the n-th
```
vtraag, line 3218: Typo:
```suggestion
          The order of the edges is the same as in the graph, so the i-th
```

src/igraph/__init__.py
```diff
@@ -3190,6 +3190,54 @@ def DataFrame(klass, edges, directed=True, vertices=None, use_vids=False):
...
+
+        df = pd.DataFrame(
+            {attr: self.vs[attr] for attr in self.vertex_attributes()},
+            index=list(range(self.vcount())),
+        )
+        if 'name' in df.columns:
+            df.set_index('name', inplace=True)
+        return df
+
+    def get_edge_dataframe(self):
```
vtraag, line 3212: Here, again, we might make it optional for the user to use either integer vertex indices, or the vertex labels (in the `source` and `target` columns).

vtraag (commented):
Thanks for the implementation @iosonofabio, I have a couple of minor corrections, and some suggestions for improving it.

src/igraph/__init__.py
iosonofabio: The order of vertices is by id, not name, so one can currently index the vertex df by id or name as usual:

```python
df.iloc[myid]
df.loc[myname]
```

Most variations are one-liners, e.g.:

```python
df.index = np.arange(df.shape[0])  # From names to ids
df.set_index(mycol, inplace=True)  # From name/id to another column
```
so it's not hard for the user to achieve exactly what they want.

Instead, I would establish a decently consistent convention between nodes and edges, e.g.

1. using vertex ids as index, then in the source/target of the edge df we use vertex ids (this is my favourite)
2. using vertex names as index (current), then the source/target of the edge df use vertex names (less good because if the vertices have no names we get integers, otherwise strings, which tends to become a mess).

A bunch of ifs are not a problem, but I feel like telling the user about these one-liners in the docstring is worth more than baby-feeding them the exact variation.

src/igraph/__init__.py
iosonofabio: yep

src/igraph/__init__.py
iosonofabio: yep

src/igraph/__init__.py
iosonofabio: as above, I'd be more in favour of educating the user via docstrings than setting up a bunch of if/else for one-liners. Especially if that involves multiindices.

src/igraph/__init__.py
iosonofabio: Hmm... yes, the current version is just me being naive about the `name` attribute.

On the other hand, I am a little worried setting up a `pd.MultiIndex` with `(source, target)` columns as index. I use `pd.MultiIndex` a fair bit in my daily work, and it's kind of a pain. Say we want to get all incoming edges into vertex `i`:

```python
df.loc[df['target'] == i]  # without multi-index
df.query('target == i')  # without multi-index
df.loc[pd.IndexSlice[:, i], :]  # with multi-index
df.loc[df.index.get_level_values('target') == i]  # with multi-index
```
To me it looks like options 1 and 2 are easier to understand and actually mention the word `target`, while option 3 uses exotic objects (`IndexSlice`), exotic constructors via square brackets, and has a hidden requirement for specification of both row and column slicers (all those colons) which can otherwise fail silently (happened before). Option 4 is easier to understand but verbose.

I should have started replying to your review here, but alas I started below:

https://github.com/igraph/python-igraph/pull/355#discussion_r558615150

so what I wrote there is that ideally we wouldn't try to baby-feed one liners to our users. Having the multiindex is such a case:

```python
df.set_index(['source', 'target'], inplace=True)  # From columns to multi-index
df.reset_index(inplace=True)  # From multi-index to column, less well known I think
```
and if they use `drop=False` they can still slice using any of the 4 options. In such cases, I'd be in favour of documenting these transformations in the docstring rather than doing them for the user.

In any case this is just a slight preference, if you are strongly opinionated I'm happy to set up the multi-index too and document the one-liner required to snap out of it.

src/igraph/__init__.py
iosonofabio: This is a good point, that related to the https://github.com/igraph/python-igraph/pull/355#discussion_r558615150 below (why on Earth did I start replying there!)

Whichever we choose I agree the `source` and `target` cols/index levels should have the same thing as used to index the vertex dataframe, so that's not consistent at the moment.

I'd be tempted to advocate using vertex ids everywhere (i.e. keep this one and revert the vertex df to an index with just ids) because we know for a fact they are unique and well formatted. If we at some point start allowing people to use any Python hashables as vertex names, they will need to be here too which I'm a little afraid might cause a mess with dtypes.

But as for this whole PR, I'm only slightly opinionated, if you feel like you are using vertex names a lot more often than ids I'm happy to switch over everything to vertex names.

src/igraph/__init__.py
iosonofabio: yep, see comment above

iosonofabio:
Hi Vincent,

Now I've tried to do the opposite of what you asked for, i.e. simplified the functions to the bare minimum but added docstrings explaining the users how they can tweak the results. I thought it's easier seen than speculated about.

If you still think offering those output variations as keyword arguments is useful, I'm happy to add them and push another commit.

src/igraph/__init__.py
ntamas: I agree with @iosonofabio here if that matters

src/igraph/__init__.py
ntamas: > If we at some point start allowing people to use any Python hashables as vertex names,

oh please don't even think about that :) One has to draw a line somewhere.

iosonofabio: We could do: anything that can be converted into a hashable without a loss of generality is accepted 😉

iosonofabio:
I rebased to have hopefully a working CI. It should still fail because the tests are not adapted to the 'ID-only' policy proposed.

ntamas:
I think that `pandas` is not installed in the CI environment, hence the failures.

iosonofabio:
Would you like me to install it on CI or how are we going about testing this and other optional dependencies?

ntamas:
First of al, you should use this to skip the test if Pandas is not installed:

```
try:
    import pandas as pd
except ImportError:
    pd = None

[... then later on in the test case ...]

    @skipIf(pd is None, "test case depends on Pandas")
    def testSomeStuffWithPandas(self):
        ...
```

Then you can add Pandas to `appveyor.yml` in the `install:` section after the `pip install cibuildwheel` step to make sure that the Pandas testcase gets a chance to execute in the CI.

iosonofabio:
Alright, I will:
- double check all is working
- rebase on top of the `2to3` PR once that is merged
- lint/fix with black
- tag "ready for review"

iosonofabio:
Alright, rebased. Tests are passed locally, let's wait for CI.

iosonofabio:
I think this is good to go now

src/igraph/__init__.py
vtraag: Ha, I actually use this somewhere: I have a graph that determines how other graphs are coupled, see https://leidenalg.readthedocs.io/en/stable/multiplex.html#slices-to-layers.

But using such graphs as indices might become a bit more complicated for a pandas dataframe. In principle, pandas dataframes can simply store any Python object, as far as I know. I am not sure if there are any limitations on this for indices though?

I think it would make more sense to start working with names as much as possible (or at least try to support that if people want to), so for me, having this as an option would be good. Of course, it should somehow be consistent, but for that reason I think it should be left to a user to specify whether he wants the index to be with names, or with vertex integer identifiers. The default option should then be consistent with each other (and probably default to integers, or perhaps default to names only if names are available?)

src/igraph/__init__.py
vtraag: Frankly, getting the source and target names are five lines of code, which I don't consider entirely trivial for end users. I personally never had too much trouble with `MultiIndex`, but maybe I just got lucky. But good if this is in the documentation.

vtraag (approved):

ntamas:
Okay, Vincent approved so here we go now! Thanks a lot!

Issue #357, opened 2021-01-16
Graph constructor from scipy.sparse adjacency matrix


iosonofabio:
**What is the feature or improvement you would like to see?**
Large, sparse graphs are often represented using `scipy.sparse` adjacency matrices. Currently `igraph.Graph.Adjacency` expects a list of lists, it would be great to teach it to load the standard scipy matrix formats.

Even a single format (e.g. COO) would be great, because one can (somewhat expensively) convert between formats in `scipy` anyway.

**Use cases for the feature**
Large, sparse graphs:

```python
adj = scipy.sparse.coo_matrix([[0, 1], [1, 0]])

# Goal
g = igraph.Graph.Adjacency(adj)

# Current
adj_list_of_lists = [list(x) for x in adj.todense()]  # <-- this can be expensive!
g = igraph.Graph.Adjacency(adj_list_of_lists)
```

**Draft implementation**
Scipy sparse matrices contain the nonzero entries in some attribute (e.g. `adj.data`). We could make an empty graph with the right number of vertices (`adj.shape[0]`) and then fill in only the edges we need, with the right weights if the matrix is not bool.

```python
if isinstance(mat, scipy.sparse.spmatrix):
    mat = mat.tocoo()
    edges = list(zip(mat.row, mat.col))  # <- this stays sparse
    g = igraph.Graph(mat.shape[0], edges)
    if mat.dtype != bool:
        g.es['weights'] = list(mat.data)   # <- also sparse
    return g
```

**References**

https://docs.scipy.org/doc/scipy/reference/sparse.html

**PS:** I'd be happy to add this in a PR.

ntamas:
This would be a great addition. There are two things to watch out for:

1. Currently `Graph.Adjacency` is implemented in the C glue code. You need to override it in the Python layer and use `super()` to fall back to the original implementation if you did _not_ receive a SciPy sparse matrix.

2. When testing for `scipy.sparse.spmatrix` with `isinstance()`, you need to import `scipy.sparse.spmatrix` lazily and handle `ImportError` exceptions gracefully so we don't introduce a hard dependency on SciPy.

iosonofabio:
Good point and yes - that's the plan. I'll write a PR then! Thx

ntamas:
[...]

PR #359, opened 2021-01-19
Generate graph from scipy sparse matrix

iosonofabio:
This tries to implement #357, i.e. generating a `Graph` object from a sparse scipy matrix, which is the de facto standard in Pythonlands for such cases.

The key is to bypass `GraphBase.Adjacency`, which dispatches to a C function expecting a dense matrix. Instead, I reproduce the logic of the various C-level functions depending on the mode (e.g. `igraph_i_adjacency_lower`, all in the `igraph` project, `src/constructors.c`).

A few notes:
- graceful try to import scipy to check if the matrix is sparse
- added a case for (dense) `numpy` 2D `ndarrays` (which include `numpy.matrix` object) with a similar graceful import
- outsourced the sparse case to a private `_` function in a new `sparse_matrix` module to avoid cluttering the main module. Happy to discuss namespacing or alternatives
- the details of the logic use a more Pythonic syntax than their C counterparts, while trying to be somewhat efficient. No efforts were undertaken to actually optimize these operations, and I wouldn't mind getting rid of some black magic if you don't like it (sums over conditional lazy comprehensions).

Funnily enough, the concrete case that prompted me to make this PR is a *weighted* adjacency sparse matrix... so I would like to implement the weighted case as well, any suggestions how to call the class method in that case? Perhaps just `graph.AdjacencyWeighted`?

No tests written yet.

edit: Oh I just saw `Weighted_Adjacency`.

### DIFF
--- a/src/igraph/__init__.py
+++ b/src/igraph/__init__.py
@@ -46,9 +46,9 @@ from igraph.summary import *
 from igraph.utils import *
 from igraph.version import __version__, __version_info__
 from igraph.sparse_matrix import (
-        _graph_from_sparse_matrix,
-        _graph_from_weighted_sparse_matrix,
-        )
+    _graph_from_sparse_matrix,
+    _graph_from_weighted_sparse_matrix,
+)
 
 import os
 import math
@@ -2195,14 +2195,20 @@ class Graph(GraphBase):
 
         if (sparse is not None) and isinstance(matrix, sparse.spmatrix):
             return _graph_from_weighted_sparse_matrix(
-                klass, matrix, mode=mode, attr=attr,
+                klass,
+                matrix,
+                mode=mode,
+                attr=attr,
             )
 
         if (np is not None) and isinstance(matrix, np.ndarray):
             matrix = matrix.tolist()
 
         return super().WeightedAdjacency(
-            matrix, mode=mode, attr=attr, loops=loops,
+            matrix,
+            mode=mode,
+            attr=attr,
+            loops=loops,
         )
 
     def write_dimacs(self, f, source=None, target=None, capacity="capacity"):
--- a/src/igraph/__init__.py
+++ b/src/igraph/__init__.py
@@ -41,6 +41,7 @@ from igraph.formula import *
 from igraph.layout import *
 from igraph.matching import *
 from igraph.operators import *
+from igraph.sparse_matrix import *
 from igraph.statistics import *
 from igraph.summary import *
 from igraph.utils import *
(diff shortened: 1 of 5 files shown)


iosonofabio:
Huh, CI fails in an interesting way and I wouldn't mind some advice. It has to do with inheritance (alas!).

This PR currently has `Adjacency` as a classmethod of `Graph`, consistent with the current declaration. If the input is not a sparse matrix, I would like to just use the old call. But it's tricky.

```python
@classmethod
def Adjacency(klass, ...):
    return GraphBase.Adjacency(...) # or return super().Adjacency(...)
```

does not work, because `GraphBase.Adjacency` is ALSO a class method, so it will construct whatever object you give it as first argument, which via the dot notation means whatever object comes before the dot. So if we call it like this it will construct a `GraphBase`, not a `Graph`. On the other hand, `return Graph.Adjacency(...)` becomes an infinite recursion.

I could make another copy of `GraphBase.Adjacency` as `GraphBase._Adjacency` and then exploit the fact that when subclassing, Python will automatically create `Graph._Adjacency`, but is that alright or do you guys know of a better method?

iosonofabio:
I pushed the `_Adjacency` hack so you guys can see it in action, but am not super happy about it, of course. Any suggestions would be welcome @ntamas @szhorvat @vtraag

ntamas:
I think that this is a bug in `GraphBase.Adjacency()`, but let me check.

ntamas:
@iosonofabio It looks like the correct syntax for calling a class method of the superclass is:

```
class InheritedGraph(Graph):
    @classmethod
    def Adjacency(cls, *args, **kwds):
        result = super(InheritedGraph, cls).Adjacency(*args, **kwds)
        return result
```

So I would try something like `super(Graph, cls).Adjacency()` from the `Graph` implementation to call `Adjacency` in `GraphBase`. Let me know if it doesn't work.

src/igraph/sparse_matrix.py
```diff
@@ -4,7 +4,7 @@
 
 from __future__ import with_statement
 
-__all__ = ()
+__all__ = ['_graph_from_sparse_matrix']
```
ntamas, line 7: Let's not export `_graph_from_sparse_matrix`; you can still import it explicitly from other modules if you want to, but it does not become a part of the declared public API if you don't add it to `__all__`. (At least that's the general consensus, but nothing in Python enforces one to treat `__all__` as the declaration of the public API).

ntamas: Also, some linters complain if `__all__` is of a mutable type (like a list) because then you could in theory import `__all__` from elsewhere and append to it. So it's better to use a tuple to make the linters shut up.

src/igraph/__init__.py
```diff
@@ -42,6 +42,7 @@
 from igraph.layout import *
 from igraph.matching import *
 from igraph.operators import *
+from igraph.sparse_matrix import *
```
ntamas, line 45: Make sure to refer to your function here explicitly if you remove it from `__all__`.

I plan to remove the wildcard imports in the long term anyway, it makes linters impossible to follow what is being imported from where and they do not print warning anymore if I mistype something while editing the file.

src/igraph/sparse_matrix.py
```diff
@@ -0,0 +1,109 @@
+# vim:ts=4:sw=4:sts=4:et
+# -*- coding: utf-8 -*-
+"""Implementation of union, disjoint union and intersection operators."""
```
ntamas, line 3: Please fix the docstring here :)

src/igraph/sparse_matrix.py
```diff
@@ -0,0 +1,109 @@
...
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc.,  51 Franklin Street, Fifth Floor, Boston, MA
+02110-1301 USA
+"""
+
+# pylint: disable-msg=W0401
+# W0401: wildcard import
+from igraph._igraph import *
```
ntamas, line 30: Let's try to adopt not using wildcard imports for new code; see my reasons in one of the other comments above. (Pylint is totally right about discouraging the use of wildcard imports).

src/igraph/sparse_matrix.py
```diff
@@ -0,0 +1,109 @@
...
+            )
+
+    if mode == ADJ_UNDIRECTED:
+        mode = ADJ_MAX
+
+    if mode in (ADJ_MAX, ADJ_MIN, ADJ_PLUS):
+        fun_dict = {
+                ADJ_MAX: max,
+                ADJ_MIN: min,
+                ADJ_PLUS: lambda *x: sum(x),
```
ntamas, line 72: I think that this function will never be called with more than two arguments, so you could write `lambda x, y: x + y` -- which is then equivalent to the `add` function from the `operator` module, so you could simply write `ADJ_PLUS: add`

ntamas (changes requested):
A few minor comments here and there, otherwise I think it's good to go!

iosonofabio:
I fixed those, but let me add the weighted case too since I'm here

src/igraph/sparse_matrix.py
iosonofabio: yep, done

src/igraph/sparse_matrix.py
iosonofabio: sounds good, done

src/igraph/sparse_matrix.py
iosonofabio: haha yes done

src/igraph/__init__.py
iosonofabio: done that

src/igraph/sparse_matrix.py
iosonofabio: yep, agreed

iosonofabio:
rebased on top of the GitHub actions PR, let's see if this works

ntamas:
@iosonofabio Can you rebase this once again on top of `master` to see if the CI checks pass? I can merge it after that.

jvail:
Sorry for spoiling your PR discussion: It is just a pity this could not make it into the latest release. Would be a very useful feature!

ntamas:
There will be a 0.9.1 soon(ish) as the API documentation build is currently broken, so I'm hoping to squeeze this into that one.

PR #360, opened 2021-01-27
CI -> Github actions

iosonofabio:
Migrating Travis CI to Github actions.

I set up two workflows:
- `build.yml` runs on all push/PR
- `develop.yml` runs on tags, supposed to make wheels and push a new release on github

I could expand `develop.yml` to push to Pypi if we wanted.

Status:
- The `build.yml` workflow runs jut fine.
- The deploy workflow is probably a little buggy.


iosonofabio:
Ok, this now passes CI on linux and OSX, and releases a version (I still have to test the tag trigger).

Why it fails on appveyor, I am unsure, they should be totally decoupled.

ntamas:
Phew, nice job! Give me a bit of time to understand it :)

Is there a way to trigger `deploy.yml` manually? I'm a bit afraid of the scenario where the "regular" build seems okay but the building of wheels fails on some combination of a platform/version. If I understand correctly, I would need to tag a commit, check whether the wheels build successfully, and if they don't, remove the tag, fix the problem and add another tag. If the workflow can be triggered manually without adding a tag, this is not a problem.

iosonofabio:
Alright, I got it to trigger manually. It goes via this thing called: `workflow_dispatch`:

https://docs.github.com/en/rest/reference/actions#create-a-workflow-dispatch-event

In theory, a button should appear in the GitHub Actions UI as shown here:

https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

but in practice it does now work on this branch. People say you must have it in the `master` branch before it shows upin the UI. Note that GitHub (probably a different team within the company) is shifting from `master` to `main` to avoid gender-biased lingo (fair enough), so I'm not sure how the two go together.

You can trigger the same event the UI would trigger using the REST API:

```bash
curl -X POST -H "Accept: application/vnd.github.v3+json" -H "Authorization: token <token>" https://api.github.com/repos/iosonofabio/python-igraph/actions/workflows/deploy.yml/dispatches -d '{"ref":"newCI"}'
```

where token must be generated from this guide:

https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token

and must include `workflow` scope.

and the last bit that I set as `"newCI"` is the branch or tag that you want to trigger. Since you are imagining triggering this **before** tagging, it would be the branch you want to trigger, most likely `master` since we'll release from there.

Bottom line:
- we can trigger the workflow via this CLI action
- probably we can just click on the UI once it's merged.
- I added instructions right at the top of the workflow file itself, to remind ourselves how to trigger when needed

The `deploy.yml` workflow does two things:
1 build wheels and source
2 make a release and upload files to it

and both are triggered only on tags or manually. This can be tuned in a few ways, but the main two ones are:

1. If you want, we could **always** build wheels on `push` (including PRs) in order to check wheels are fine, and rely on tags (or just manual trigger) only for the release creation.
2. It currently makes wheels serially, i.e. it takes some 20 minutes to make wheels for all linux architectures (same for OSX). If we wanted, we could parallelize that. Personally, I don't think it's super important because we run that workflow only once in a blue moon.

ntamas:
Awesome, thanks. Let me know when this is ready to merge -- I'm eager to try it on the master branch.

iosonofabio:
Ok, now it releases assets correctly as a GitHub release upon tagging `*.*.*` or via manual triggering:

https://github.com/iosonofabio/python-igraph/releases/tag/untagged-c6c268a6b7dc0d12a556

I am not sure how you want to coordinate this with Pypi, but I'd say since I just rebased let's merge and refine later on.

Currently unsure why AppVeyor fails.

ntamas:
AppVeyor is unrelated, it also fails in master. It is due to some Windows DLL import / export issue that I am not familiar with and that I need to take the time to understand before the next release.

PR #363, opened 2021-02-05
Remove star imports

iosonofabio:
As in the title, I know @ntamas has been thinking about this anyway

edit: second commit is just rebasing

I think failures have to do with API changes in the C core more than anything? same applies to #364


ntamas:
Thanks a lot! A few imports were missing, but it's good to go now.

Issue #362, opened 2021-02-04
Add python wrapper for the new harmonic centrality functions


Silmathoron:
Harmonic centrality has been implemented at the C level (see [this thread](https://igraph.discourse.group/t/user-defined-distance-for-unconnected-node-in-closeness-and-harmonic-centrality/183/11)) and will be part of the next release so it would be nice if the associated wrapper could be included in the python release.
You can ping me if you need reviewers.

iosonofabio:
Hi @Silmathoron, fellow Tuebingen dweller, I did my PhD at the MPI for Dev Bio next door!

I've make a PR #364 

Cheers,
Fabio

Silmathoron:
Hi there and thanks for the PR :)
I'd love to help with the review but I'm afraid I'm unable to install this branch with the `develop` branch of igraph... I get a bunch of errors about ``igraph_harmonic_centrality`` (incompatible types for arguments 3 and 5 and too many arguments)...

iosonofabio:
Don't worry too much, we are in the middle of releasing an API-breaking update on the C core of igraph, so it'll take a few days to sort this one out. We'll get it tested on CI and once that works you can clone the master branch and test it yourself.

ntamas:
#364 is merged into master now.

PR #364, opened 2021-02-05
Add harmonic centrality

iosonofabio:
As in #362


src/_igraph/graphobject.c
```diff
@@ -12902,6 +12981,39 @@ struct PyMethodDef igraphmodule_Graph_methods[] = {
...
+   "harmonic_centrality(vertices=None, mode=ALL, cutoff=None, weights=None,\n"
+   "          normalized=True)\n\n"
+   "Calculates the harmonic centralities of given vertices in a graph.\n\n"
+   "The harmonic centerality of a vertex measures how easily other\n"
+   "vertices can be reached from it (or the other way: how easily it\n"
+   "can be reached from the other vertices). It is defined as the\n"
+   "number of vertices minus one times the sum of one divided by \n"
+   "the length of each geodesic from/to the given vertex.\n\n"
+   "If the graph is not connected, and there is no path between two\n"
+   "vertices, the number of vertices is used instead the length of\n"
```
szhorvat, line 12996: "the number of vertices used" is incorrect (seems like a leftover from the old way closeness was calculated).

When there is no path between two vertices, their inverse distance is taken to be zero.

src/_igraph/graphobject.c
```diff
@@ -12902,6 +12981,39 @@ struct PyMethodDef igraphmodule_Graph_methods[] = {
...
+  {"harmonic_centrality", (PyCFunction) igraphmodule_Graph_harmonic_centrality,
+   METH_VARARGS | METH_KEYWORDS,
+   "harmonic_centrality(vertices=None, mode=ALL, cutoff=None, weights=None,\n"
+   "          normalized=True)\n\n"
+   "Calculates the harmonic centralities of given vertices in a graph.\n\n"
+   "The harmonic centerality of a vertex measures how easily other\n"
+   "vertices can be reached from it (or the other way: how easily it\n"
+   "can be reached from the other vertices). It is defined as the\n"
+   "number of vertices minus one times the sum of one divided by \n"
+   "the length of each geodesic from/to the given vertex.\n\n"
```
szhorvat, line 12994: This is a bit hard to read for me ... I would say that the harmonic centrality of a vertex is the mean _inverse distance_ to/from all other vertices.

src/_igraph/graphobject.c
```diff
@@ -12902,6 +12981,39 @@ struct PyMethodDef igraphmodule_Graph_methods[] = {
...
+   "the geodesic. This is always longer than the longest possible\n"
+   "geodesic.\n\n"
+   "@param vertices: the vertices for which the harmonic centrality must\n"
+   "  be returned. If C{None}, uses all of the vertices in the graph.\n"
+   "@param mode: must be one of L{IN}, L{OUT} and L{ALL}. L{IN} means\n"
+   "  that the length of the incoming paths, L{OUT} means that the\n"
+   "  length of the outgoing paths must be calculated. L{ALL} means\n"
+   "  that both of them must be calculated.\n"
+   "@param cutoff: if it is an integer, only paths less than or equal to this\n"
+   "  length are considered, effectively resulting in an estimation of the\n"
```
szhorvat, line 13006: Please do not suggest that the purpose of the range-limited version is estimation. I would avoid the word 'estimation' here.

src/_igraph/graphobject.c
```diff
@@ -12902,6 +12981,39 @@ struct PyMethodDef igraphmodule_Graph_methods[] = {
...
+   "  harmonic centrality for the given vertices (which is always an underestimation of the\n"
+   "  real centrality, since some vertex pairs will appear as disconnected even\n"
+   "  though they are connected).. If C{None}, the exact centrality is\n"
+   "  returned.\n"
+   "@param weights: edge weights to be used. Can be a sequence or iterable or\n"
+   "  even an edge attribute name.\n"
+   "@param normalized: Whether to normalize the raw centrality scores by\n"
```
szhorvat, line 13013: The description of normalization seems incorrect (see the C docs).

szhorvat:
@iosonofabio I looked through the docs only, and added a few notes. I know it's not finished yet.

src/_igraph/graphobject.c
iosonofabio: yep, done

src/_igraph/graphobject.c
iosonofabio: yep, done

src/_igraph/graphobject.c
iosonofabio: sure, that was the old closeness text so I just removed it now

szhorvat: We need to fix up the description of closeness in the python-igraph docs, as we also changed how closeness works recently.

src/_igraph/graphobject.c
iosonofabio: It seemed fine to me already, but I now copy-pasted and adapted the C function description.

iosonofabio:
thanks, fixed those. I guess I should implement a test

src/_igraph/graphobject.c
```diff
@@ -12902,6 +12981,34 @@ struct PyMethodDef igraphmodule_Graph_methods[] = {
...
+   "mean inverse distance to all other vertices.\n\n"
+   "If the graph is not connected, and there is no path between two\n"
+   "vertices, the inverse distance is taken to be zero.\n\n"
+   "@param vertices: the vertices for which the harmonic centrality must\n"
+   "  be returned. If C{None}, uses all of the vertices in the graph.\n"
+   "@param mode: must be one of L{IN}, L{OUT} and L{ALL}. L{IN} means\n"
+   "  that the length of the incoming paths, L{OUT} means that the\n"
+   "  length of the outgoing paths must be calculated. L{ALL} means\n"
+   "  that both of them must be calculated.\n"
+   "@param cutoff: if it is an integer, only paths less than or equal to this\n"
```
szhorvat, line 13002: There is no requirement for this to be an integer (we can have weighted graphs, and the `<= cutoff` considered rule is valid even for unweighted)

src/_igraph/graphobject.c
```diff
@@ -12902,6 +12981,34 @@ struct PyMethodDef igraphmodule_Graph_methods[] = {
...
+   "  length are considered.\n"
+   "@param weights: edge weights to be used. Can be a sequence or iterable or\n"
+   "  even an edge attribute name.\n"
+   "@param normalized: Whether to normalize the result. If True, the \n"
+   "  result is the mean inverse path length to other vertices, i.e. it \n"
+   "  is normalized by the number of vertices minus one. If False, the \n"
+   "  result is the sum of inverse path lengths to other vertices.\n"
+   "@return: the calculated harmonic centrality in a list\n"},
```
szhorvat, line 13010: "centralities" (plural)

tests/test_structural.py
```diff
@@ -355,6 +355,37 @@ def testClosenessCentrality(self):
...
+        cl = g.harmonic_centrality()
+        cl2 = [1.0] + [(1.0 + 1 / 2 * 3) / 4] * 4
+        for idx in range(g.vcount()):
+            self.assertAlmostEqual(cl[idx], cl2[idx], places=3)
+
+        g = Graph.Star(5)
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
+            cl = g.harmonic_centrality(cutoff=1.0)
+        cl2 = [1.0, 1.0, 1.0, 1.0, 1.0]
```
szhorvat, line 369: I'm a bit confused here. Does this test asserts that with `cutoff=1.0`, all harmonic centralities for the star graph should be 1?

If so, that is incorrect. We should have `[1.0, 0.25, 0.25, 0.25, 0.25]`, as we take paths of length 1, but not not longer. Thus each of the "spoke nodes" can reach only the centre node, so the inverse distances of one spoke node are [1,0,0,0] which averages to 0.25. This is what I get with the Mathematica interface.

What am I misunderstanding? While I did fix some issues 9 days ago, those already seem to be picked up by python-igraph.

szhorvat: image

src/_igraph/graphobject.c
```diff
@@ -4118,6 +4118,85 @@ PyObject *igraphmodule_Graph_closeness(igraphmodule_GraphObject * self,
...
+      igraph_vector_destroy(&res);
+      if (weights) { igraph_vector_destroy(weights); free(weights); }
+      igraphmodule_handle_igraph_error();
+      return NULL;
+    }
+  } else if (PyNumber_Check(cutoff)) {
+    PyObject *cutoff_num = PyNumber_Float(cutoff);
+    if (cutoff_num == NULL) {
+      igraph_vs_destroy(&vs); igraph_vector_destroy(&res);
+      return NULL;
```
ntamas, line 4173: `weights` should be destroyed here. It might also be a bug in the closeness centrality if this part of the code was copied from there.

ntamas: Just checked; the problem also appears there (but not for betweenness or edge betweenness).

tests/test_structural.py
ntamas: Indeed; this was also confirmed by the fact that the test case failed. I've updated the test case.

ntamas:
This seems to be good to go; @iosonofabio is there anything else that you would like to add or can I merge this?

iosonofabio:
Hi ahead thank you! Swamped by grants...

PR #371, opened 2021-03-05
Add matplotlib plotting to VertexClustering

### LINKED ISSUE #370: Plotting VertexClustering fails when targeting matplotlib
**Describe the bug**
Ordinarily, one could plot a `VertexClustering`, for example using

```python
G = ig.Graph.Famous('Zachary')
comms = G.community_leiden('modularity')
ig.plot(comms)
```
image

However, when doing `ig.plot(comms, target=ax)`, where `ax` is a `matplotlib` `Axes`, this throws the error
`AttributeError: 'VertexClustering' object has no attribute 'layout'`.

**To reproduce**
```python
import igraph as ig
G = ig.Graph.Famous('Zachary')
comms = G.community_leiden('modularity')
fig, ax = plt.subplots()
ig.plot(comms, target=ax)
```

**Version information**
`python-igraph` version 0.9.0, linked against external shared `igraph` library version 0.9.0-135-g4019b1f29.

iosonofabio:
This PR aims to fix #370 

ATM there is no specific test for this.

### DIFF
--- a/src/igraph/drawing/graph.py
+++ b/src/igraph/drawing/graph.py
@@ -983,6 +983,8 @@ class MatplotlibGraphDrawer(AbstractGraphDrawer):
         from matplotlib.patches import FancyArrowPatch
         from matplotlib.patches import ArrowStyle
         import numpy as np
+        from collections import defaultdict
+        from igraph.clustering import VertexClustering
 
         def shrink_vertex(ax, aux, vcoord, vsize_squared):
             """Shrink edge by vertex size"""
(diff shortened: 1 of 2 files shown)


ntamas:
It's hard to test anyway, so let's just roll with the implementation. Let me know when it's ready for review.

iosonofabio:
The whole "mark_groups" option is a little weird when used in conjunction with a `VertexClustering` as the first argument of `igraph.plot`, but changing that breaks API for the Cairo backend, so let's not touch it here.

Ready for merge.

vtraag:
> The whole "mark_groups" option is a little weird when used in conjunction with a `VertexClustering` as the first argument of `igraph.plot`, but changing that breaks API for the Cairo backend, so let's not touch it here.

Well, yes, sort of. A simple boolean would suffice, simply to indicate whether you want to draw the convex hulls or not.

iosonofabio:
Fair enough, I didn't know that was an option since it's not mentioned in the docs. I added it and also cleaned up the code a little, and added this info to the docs.

I then rebased on top of the current master, so if it passes CI it should be good to go.

src/igraph/drawing/graph.py
```diff
@@ -1010,6 +1012,61 @@ def callback_edge_offset(event):
...
+
+        # mark groups: the used data structure is eventually the dict option:
+        # tuples of vertex indices as the keys, colors as the values. We
+        # convert other formats into that one here
+        if "mark_groups" in kwds:
+            if kwds["mark_groups"] is False:
+                del kwds["mark_groups"]
+            elif (kwds["mark_groups"] is True) and (clustering is not None):
+                pass
+            # Deferred import to avoid a cycle in the import graph
```
ntamas, line 1031: The deferred import is not here but above

src/igraph/drawing/graph.py
```diff
@@ -1010,6 +1012,61 @@ def callback_edge_offset(event):
...
+                    membership = [x[0] for x in membership]
+                clusters = sorted(set(membership))
+                n_colors = len(clusters)
+                cmap = mpl.cm.get_cmap("viridis")
+                colors = [cmap(1.0 * i / n_colors) for i in range(n_colors)]
+                c = [colors[clusters.index(i)] for i in membership]
+                kwds["vertex_color"] = c
+
+                # mark_groups if not explicitely marked
+                if ("mark_groups" in kwds) and (kwds["mark_groups"] is True):
```
ntamas, line 1061: `if kwds.get("mark_groups") is True` is enough; `.get()` returns `None` if the key is not in the dict

ntamas (approved):
Okay, seems good, only a few minor changes were requested but I'll do these myself.

PR #375, opened 2021-03-12
Add GraphML support for Windows wheels

vtraag:
This PR links against the `libxml2` from `vcpkg` in order to provide GraphML support on Windows. This is implemented under GitHub Actions, instead of AppVeyor. Once successful, AppVeyor can be removed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,8 +12,6 @@ on:
     tags-ignore:
       - '*.*.*'
 
-
-
 jobs:
   build_linux:
     runs-on: ubuntu-latest
@@ -83,3 +81,50 @@ jobs:
       env:
         TESTING_IN_TOX: 1
 
+  build_windows:
+    runs-on: windows-latest
+    env:
+      IGRAPH_CMAKE_EXTRA_ARGS: -DVCPKG_TARGET_TRIPLET=x64-windows-static-md -DCMAKE_TOOLCHAIN_FILE=%VCPKG_INSTALLATION_ROOT%/scripts/buildsystems/vcpkg.cmake
+
+    strategy:
+      matrix:
+        python-version:
+        - 3.6
+        - 3.7
+        - 3.8
+        - 3.9
+        - pypy-3.7
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Init C core submodule
+      run: |
+        git submodule update --init
+        cat vendor/source/igraph/.git
+        cat .git/modules/vendor/source/igraph/HEAD
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Cache VCPKG
+      uses: actions/cache@v2
+      with:
+        path: $(VCPKG_INSTALLATION_ROOT)\installed
+        key: vcpkg-installed-${{ runner.os }}
+    - name: Install OS dependencies
+      run: |
+        choco install winflexbison3 ninja
+        %VCPKG_INSTALLATION_ROOT%\vcpkg.exe integrate install
+        %VCPKG_INSTALLATION_ROOT%\vcpkg.exe install libxml2:x64-windows-static-md
+    - name: Install Python dependencies
+      run: |
+        # Pypi has no pip by default, and ubuntu blocks python -m ensurepip
+        # However, Github runners are supposed to have pip installed by default
+        # https://docs.github.com/en/actions/guides/building-and-testing-python
+        #wget -qO- https://bootstrap.pypa.io/get-pip.py | python
+        python -m pip install --upgrade pip
+        pip install tox tox-gh-actions
+    - name: Test with tox
+      run: tox
+      env:
+        TESTING_IN_TOX: 1


vtraag:
The `igraph` C core now builds on Windows on GitHub Actions, linking against the `libxml2` dependency from `vcpkg`. However, for some reason, some of the symbols cannot be found when compiling the `python-igraph` C core. Since it is all statically linked, I would have assumed we would not encounter any missing `xml` symbols due to misconfigured paths or anything. Thoughts on what might be happening here? See https://github.com/igraph/python-igraph/pull/375/checks?check_run_id=2096393567#step:8:1301 for details.

vtraag:
OK, this now finally seems to work, at least on Python 3.6 and 3.7 by bundling the dynamic libraries using `delvewheel`. I have tested this also locally on my Windows machine, and I can confirm that I am able to read GraphML files (there is no unit test for this yet).

On Python 3.8 this does not work unfortunately, because it depends on the newly introduced `os.add_dll_directory(libs_dir)`. That is, when doing `import igraph`, I get the infamous `ImportError: DLL load failed while importing _igraph: The specified module could not be found.`. For some reason this does seem to work on the CI, but it does not work on my Windows machine locally. Loading the DLLs explicitly by using `WinDLL` from `ctypes` (the mechanism being used pre-3.8) does work, also on Python 3.8.

I previously tried to get static linking to work, which had some separate issues on its own. I will now retry to do that.

vtraag:
If someone does have a suggestion for getting `os.add_dll_directory` to work properly, that might be workable as well. I am just not sure how robust the mechanism is.

vtraag:
OK, it seems that the static build now works correctly on all versions of Python that we support (>= 3.6). There are now still a couple of things to do:

- [ ] Include unit test for loading GraphML
- [ ] Build 32-bit wheels, with GraphML support
- [ ] Cache `vcpkg`

Additionally, because the C core does not change so much, I thought we could also cache the `vendor/install` directory. That should reduce the build time significantly.

Perhaps it would also be good to clean up `setup.py` a little bit and document somewhere how it can be build on Windows. What place would be best suited for that?

Everything is now build on GitHub actions, so we can remove the AppVeyor build (this PR already removes `appveyor.yml`, but it is still triggered somehow). In this PR I now have gotten rid of `tox` in favour of testing the binary wheels themselves directly. That way, we are sure that whatever we are shipping is actually tested. I think this is a more robust test strategy.

I think we could therefore also get rid of the separate Deploy GitHub Action workflow (`deploy.yml`). I believe that this workflow uploads the binary wheels to a new release whenever a commit is tagged with a `x.x.x` version. In the current workflow, it uploads the binary wheels to PyPI when a release is made. As far as I am concerned, we don't need to include the binary wheels separately in the GitHub release, only having them on PyPI is sufficient I think. But @ntamas and @iosonofabio, I know you previously discussed about uploading things to PyPI, so feel free to chime in here.

Finally, do we want to have debug builds for regular builds and release builds upon, well, release? We could use `RelWithDebInfo`, as we now also do in the `igraph` repository, for building the C core. However, I don't know how to tell the Python build to make a debug build or a release build. Do you @ntamas ? We can set `C(XX)FLAGS`, which would hence need to be set separately for all Windows and Linux/macOS, but I'm not sure if that would be the normal route to go for Python.

adang1345:
Regarding your observation that `os.add_dll_directory` isn't working: could the problem be related to https://github.com/conda-forge/python-feedstock/issues/444?

vtraag:
@adang1345, yes, that could very well be the source of the problem. Hopefully Anaconda will address the issue soon enough! It will happen often enough that people still end up installing something through `pip`, even if it is available through `conda`.

At the moment, the build is working fine with static linking, alleviating the need for bundling any dynamic dependencies. The only remaining dynamic dependency is `ws_32`, but I believe we can safely assume it to be present on all systems with Windows.

ntamas:
> Additionally, because the C core does not change so much, I thought we could also cache the vendor/install directory. That should reduce the build time significantly.

I don't know how caching works in Github Actions -- can we ensure that the cache is cleared whenever the commit reference that `vendor/source/igraph` points to changes? Or does the cache have to be invaliated manually?

Regarding the deployment workflow: the idea was to upload a new release to PyPI automatically for each tagged release. Having the wheels on the Github releases page was a "side effect" -- Appveyor uploaded the wheels to the Github release automatically. I think if we get rid of Appveyor then this will be solved as well, and the wheels would end up on PyPI only. I've checked how other Python projects do and there seems to be no clear consensus; out of the four big Python projects I checked (NumPy, SciPy, Matplotlib, Pandas) SciPy posted the wheels on the Github releases page but the other three did not. I'm okay with not having them on the Github releases page.

> However, I don't know how to tell the Python build to make a debug build or a release build. 

I don't think that you can turn this on or off easily. You could pass extra CMake switches to force a `RelWithDebInfo` build of the C core (but then that messes up the caching of `vendor/install/igraph`), but it will not affect the glue code between Python and C. As far as I know, you would need a debug build of Python itself to get any useful debug info for extensions. For instance, on Ubuntu you can install `pythonX.Y-dbg` to get a debug build of the Python interpreter (version X.Y). I'd say let's leave it for the time being as is, and if we really badly need a debug build to track down some problem then we can return to it.

vtraag:
Ha, you were a bit quicker with merging than I anticipated. :) I will add some of the other stuff separately then.

> I don't know how caching works in Github Actions -- can we ensure that the cache is cleared whenever the commit reference that `vendor/source/igraph` points to changes? Or does the cache have to be invaliated manually?

Yes, we can ensure that the cache is always cleared when the commit reference changes.

> Regarding the deployment workflow: the idea was to upload a new release to PyPI automatically for each tagged release.
> I'm okay with not having them on the Github releases page.

OK, I can also do the automatic upload to PyPI on tagging, not on releasing. I will have to change it.

> > However, I don't know how to tell the Python build to make a debug build or a release build.
> I'd say let's leave it for the time being as is, and if we really badly need a debug build to track down some problem then we can return to it.

OK, that is definitely easier, and then we are also sure that the Release is being tested. Is there any downside to using `RelWithDebInfo` vis-a-vis `Release` builds? If not, building by default with `RelWithDebInfo` might provide more useful debugging feedback from users.

vtraag:
> Perhaps it would also be good to clean up `setup.py` a little bit and document somewhere how it can be build on Windows. What place would be best suited for that?

Could you suggest where I should document linking to `libxml2` on Windows @ntamas? Should I simply include this in the `README.md`?

PR #377, opened 2021-03-25
Added unittest for GraphML

vtraag:
This PR adds a unit test to test reading GraphML. The test is skipped if the functionality is not available from the C core.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -259,6 +259,7 @@ class IgraphCCoreCMakeBuilder(IgraphCCoreBuilder):
         # Finally, add the source folder path
         args.append(str(build_to_source_folder))
 
+        print("Running {}".format(" ".join(args)))
         retcode = subprocess.call(args)
         if retcode:
             return False
--- a/setup.py
+++ b/setup.py
@@ -259,7 +259,6 @@ class IgraphCCoreCMakeBuilder(IgraphCCoreBuilder):
         # Finally, add the source folder path
         args.append(str(build_to_source_folder))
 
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -259,6 +259,7 @@ def compile_in(self, source_folder, build_folder, install_folder):
         # Finally, add the source folder path
         args.append(str(build_to_source_folder))
 
+        print("Running {}".format(" ".join(args)))
```
ntamas, line 262: Can we live without this line or did you need this line often during debugging? (I can leave it in if it was needed so often that you'd rather have it in the script).

tests/test_foreign.py
```diff
@@ -339,9 +399,6 @@ def testPickle(self):
             self.assertTrue(isinstance(g, Graph))
             self.assertTrue(g.vcount() == 3 and g.ecount() == 1 and not g.is_directed())
             g.write_pickle(tmpfname)
-
-    
-
```
ntamas, line 344: Some separation would be great between test cases. I agree that three lines is maybe too much, but no newlines at all may be too little :)

setup.py
vtraag: No, we can remove it. I just added it for debugging, but I didn't thought it hurt to be explicit about it.

tests/test_foreign.py
vtraag: Ha, oh sorry, must have removed those lines accidentally :)

ntamas (approved):

ntamas:
Thanks!

PR #378, opened 2021-03-25
CI: Updated

vtraag:
This PR improves the current CI setup. In particular, I
- Added Windows 32-bit build.
- Improved caching.
- Removed old workflows and test setup.

There is one final thing regarding the PyPI upload, I did not yet test this. We probably first need to create a token for this and test it against the test PyPI. @iosonofabio , @ntamas , you already set this up previously using tokens or not? I just wanted to already share this already.


ntamas:
We never tested the automatic upload to PyPI "live" because there was only one release since this was added by @iosonofabio and I screwed something up with the setup. But I feel more comfortable uploading things from my own computer with `twine` after having downloaded the wheels from the CI anyway.

vtraag:
It's also fine by me to simply remove the automatic upload then. You can simply download all wheels from the build and upload them.

tox.ini
```diff
@@ -1,31 +0,0 @@
...
-deps =
-    scipy; platform_python_implementation != "PyPy"
-    numpy; platform_python_implementation != "PyPy"
-    networkx; platform_python_implementation != "PyPy"
-    pandas; platform_python_implementation != "PyPy"
-passenv = PATH
-setenv =
-    TESTING_IN_TOX=1
-
-[flake8]
```
ntamas, line 28: Can we leave the `[flake8]` section in `tox.ini`? This does not really belong to `tox` itself; these are settings for the linter that I use (`flake8`). For historical reasons, `flake8` is configured from `tox.ini`.

vtraag: It seems that you can also put them specifically in a `.flake8` file, https://flake8.pycqa.org/en/latest/user/configuration.html#configuration-locations. Perhaps that clarifies the intention of this file better?

vtraag:
I now noticed that I forgot to include the macOS as a `need` for the release. If we remove the automatic release this doesn't matter anyway, but I just wanted to note it here.

vtraag:
OK, I removed the PyPI upload. Apart from the `.flake8` file, this is good to go as far as I am concerned.

tox.ini
ntamas: I'll add a `.flake8` file and then merge.

Issue #343, opened 2020-11-11
igraph_tree_game() should be exposed in the Python interface


ntamas:
**What is the feature or improvement you would like to see?**

`igraph_tree_game()` from the C library should be exposed in the Python interface.

**Use cases for the feature**

Well, to generate random trees :)

ntamas:
As requested by @szhorvat in #340

ntamas:
Implemented by #385

PR #385, opened 2021-04-18
Random tree games

iosonofabio:
Addressing  #343.

This raises an interesting point: I now made two new global variables in the igraph namespace for this: `RANDOM_TREE_PRUFER` and `RANDOM_TREE_LERW`: probably that's not a great idea? Perhaps we should check strings as parameters for the Python function instead?

e.g. 

```python

igraph.Graph.Random_Tree(10, False, igraph.RANDOM_TREE_PRUFER) # current
igraph.Graph.Random_Tree(10, False, "Prufer") # perhaps better?
```

That applies to a number of other situations, of course, so I'm wondering what is the better choice, see also #386

### DIFF
--- a/src/_igraph/graphobject.c
+++ b/src/_igraph/graphobject.c
@@ -674,7 +674,7 @@ PyObject *igraphmodule_Graph_delete_edges(igraphmodule_GraphObject * self,
 }
 
 /**********************************************************************
- * tructural properties                                              *
+ * structural properties                                              *
  **********************************************************************/
 
 /** \ingroup python_interface_graph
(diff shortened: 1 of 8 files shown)


ntamas:
The way it typically works is that you add a conversion function in `src/convert.c` that converts from strings to the appropriate enum members, and then you use those enum members instead. I started moving away from global integer variables a while ago, although of course I cannot remove any existing stuff in order not to break backwards compatibility.

src/_igraph/graphobject.c
```diff
@@ -3400,6 +3400,60 @@ PyObject *igraphmodule_Graph_Tree(PyTypeObject * type,
...
+                                  PyObject * args, PyObject * kwds)
+{
+  long int n, directed_o, tree_method_o;
+  igraphmodule_GraphObject *self;
+  igraph_t g;
+
+  static char *kwlist[] = { "n", "directed", "method", NULL };
+
+  if (!PyArg_ParseTupleAndKeywords(args, kwds, "lll", kwlist,
+                                   &n, &directed_o, &tree_method_o))
```
ntamas, line 3427: `directed` is a boolean so I'd rather go for `PyObject* directed_o` and use `PyObject_IsTrue(...)` for sake of consistency with how it's done elsewhere in the library.

As for the `method` argument, look into `convert.c` to see how it's typically done for other enums.

ntamas (changes requested):
Thanks a lot for the contribution! It looks good to me in general, except the handling of arguments, see my comments there.

src/_igraph/graphobject.c
iosonofabio: alright let me take a look

iosonofabio: yep I've changed this now.

iosonofabio:
I rebased on top of the realize_degree_sequence commits, changed this to use the strict-ish conversion function, and all tests pass, so I think it's ready for merging.

ntamas (approved):

ntamas:
Thanks! I'll add default argument handling and then it's good to go.

szhorvat:
It'd be nice to have consistent capitalization for method names, i.e. `Prufer`, `LERW` or `prufer`, `lerw`. Now it's mixed.

I recommend Prufer as default unless the graph is directed. There is no difference in the distribution with the two methods, only in the performance.

ntamas:
Matching is case-insensitive, the docstrings consistently use lowercase everywhere, the tests use uppercase here and there to test the cast insensitiveness.

I used LERW as the default because then I don't need to make the default dependent on another variable. How big is the performance difference? If it is significant, I'll change the default.

szhorvat:
Would have to tes, but I think not significant.

PR #386, opened 2021-04-19
Realize degree sequence

### LINKED ISSUE #316: Expose realize_degree_sequence
This is a feature request to expose `igraph_realize_degree_sequence` in the Python interface.

It's best to do this after https://github.com/igraph/igraph/pull/1445 is merged.

iosonofabio:
Trying to fix #316 

This also relates to #385. In that PR, I added two constants to the global API of `python-igraph`, mimicking what happens at the C level. In this case, I used strings as options in Python, which keeps the API cleaner. I think both PR should use the same approach, but I wanted you guys to see the options.

I'd prefer to use strings as in here, any other opinions @ntamas @szhorvat @vtraag ?

### DIFF
--- a/src/_igraph/graphobject.c
+++ b/src/_igraph/graphobject.c
@@ -2804,7 +2804,6 @@ PyObject *igraphmodule_Graph_Realize_Degree_Sequence(PyTypeObject *type,
   igraph_edge_type_sw_t allowed_edge_types;
   igraph_realize_degseq_t method;
   PyObject *outdeg_o, *indeg_o, *edge_types_o, *method_o;
-  PyObject *repr;
   igraphmodule_GraphObject *self;
   igraph_t g;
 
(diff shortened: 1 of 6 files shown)


ntamas:
Strings for the win, no question about that, but see `convert.c` for examples about converting strings to igraph enums.

iosonofabio:
Thanks @ntamas. I've now added two functions in `convert.h` and `convert.c` for this and used them like the neighboring `igraphmodule_PyObject_to_pagerank_algo_t` is used. But now it fails to raise `ValueError` in the tests...

ntamas:
I've made a dubious decision when implementing `igraphmodule_PyObject_to_enum` -- it accepts partial matches as well if there is no exact match. So, if you have a unique enum member that starts with `s` (like `smallest`), and you pass `should_fail` as a string to the conversion routine, igraph will happily match `should_fail` to `smallest` because a prefix of length 1 matches and there is no other enum member with an equally long or longer match. Yes, I know, it's horrible. We can kill it in another PR.

iosonofabio:
haha, yes it's horrible. Can I add another function that does a strict conversion check? Perhaps that'd be easiest

ntamas:
Sure, go ahead!

iosonofabio:
Alright, I've added a function called `igraphmodule_PyObject_to_enum_strict`. Is this name alright or would you prefer a different one?

Also, in the `_strict` function I got rid of the part that makes lowercase out of everything. Or did you want to keep that?

Cheers,
Fabio

ntamas:
The name is okay. I'd rather keep the case-insensitiveness (but then it's not so strict any more, is it? :) ).

iosonofabio:
We could call it `igraphmodule_PyObject_to_enum_strictish`... or we could use the "SuperHexagon" style: "_strict", "_stricter", "_strictest".

If this passes CI it's ready for merging, and then I'll rebase the other one on top of this and use the "strict" function there as well.

szhorvat:
in case it influences your design, there will be another function for the bipartite case. see `is_bigraphical` for reference

iosonofabio:
This is ready to merge, then I can move on to the random trees that will use the strict function defined here.

src/_igraph/graphobject.c
```diff
@@ -2791,6 +2791,67 @@ PyObject *igraphmodule_Graph_LCF(PyTypeObject *type,
...
+
+  igraph_vector_t outdeg, indeg;
+  igraph_vector_t *indegp = 0;
+  igraph_edge_type_sw_t allowed_edge_types;
+  igraph_realize_degseq_t method;
+  PyObject *outdeg_o, *indeg_o, *edge_types_o, *method_o;
+  igraphmodule_GraphObject *self;
+  igraph_t g;
+
+  static char *kwlist[] = { "outdeg", "indeg", "allowed_edge_types", "method", NULL };
```
ntamas, line 2810: The keyword argument ordering is a bit tricky for this function. I agree with your initial assessment that `allowed_edge_types` and `method` should be mandatory; after all, the purpose of this function is to let the user have fine-grained control over how the graph generation happens. However, it bothers me that as a consequence, `indeg` has to be set to `None` explicitly for undirected graphs.

Therefore, I'm leaning towards providing a default value for all the keyword arguments except `outdeg`. I'll do it in a separate commit, just wanted to mention it here in case you disagree.

src/_igraph/graphobject.c
```diff
@@ -12605,6 +12666,26 @@ struct PyMethodDef igraphmodule_Graph_methods[] = {
...
 
+  {"Realize_Degree_Sequence", (PyCFunction) igraphmodule_Graph_Realize_Degree_Sequence,
+    METH_VARARGS | METH_CLASS | METH_KEYWORDS,
+    "Realize_Degree_Sequence(outdeg, indeg, allowed_edge_types, method)\n--\n\n"
+    "Generates a graph from a degree sequence.\n\n"
+    "@param outdeg: the degree sequence of an undirected graph (if indeg=None), "
+    "or the out-degree sequence of a directed graph.\n"
+    "@param indeg: None to generate an undirected graph, the in-degree sequence "
+    "to generate a directed graph.\n"
+    "@param allowed_edge_types: for directed graphs, only 'simple_sw' is currently \n"
```
ntamas, line 12677: I'd rather not provide details here; if we update the C core and add other implementations, there's a 99% chance that we forget to update the documentation here. I'll remove the details in a separate commit and just mention that we support whatever the C core supports.

ntamas (approved):

src/_igraph/graphobject.c
iosonofabio: I agree with you, had the same doubt, whatever you choose is going to be just fine for me

szhorvat: As defaults, I recommend "simple graph" and "smallest first" method. Simple graphs are the least trivial to construct and the most often needed. "Smallest first" produces connected graphs in the undirected case, which is nice (although everyone else implements "largest first" only).

ntamas: @szhorvat I read your mind yesterday; this is exactly how I set up the defaults.

Issue #389, opened 2021-04-27
function parity with igraph C core


iosonofabio:
We should aim, as a continuous self-improvement kind of goal, to close the gap in function availability between the igraph C core and the Python interface. To help with that, let's try and keep track of what functions:
- are missing from the Python interface
- would be useful to have there
- are already implemented in the C core

Other igraph folks: please feel free to edit the list adding functions you find or deleting them if they got implemented or if you think they aren't necessary in the Python interface:

- [ ] `maximal_cliques_count`
- [ ] `maximal_cliques_subset`
- [ ] `clique_size_hist` and other histogram functions (useful for large graphs, where a list of all items would be too large)
- [ ] weighted cliques functions (?)
- [ ] `vertex_coloring_greedy`
- [ ] `trussness`
- [ ] `eigen_laplacian`
- [ ] `laplacian_spectral_embedding`
- [ ] SIR simulations (do we want it?)
- [ ] `is_eulerian` and related functions
- [ ] `is_bigraphical`
- [ ] graphlet functions
- [ ] `is_matching` and `is_maximal_matching` (they seem to be there partially, perhaps a quick fix?)
- [ ] `maximum_bipartite_matching` (it's there but buggy)
- [ ] ...

szhorvat:
I am not completely comfortable with the "graphlet" functions. I looked into the paper, and could never quite make sense of what igraph's function returns. Finally I decided it's not worth spending the time on as it is very niche functionality. This use of the word "graphlet" is also an outlier. Most commonly, "graphlet" is used as an alternative name for motifs (what igraph calls "motifs"). This use is entirely different.

Given that I am not 100% convinced that the results are even correct, I am considering removing graphlets from IGraph/M. (But I'd be happy if someone could help me make sense of the results.)

https://igraph.discourse.group/t/on-the-naming-of-igraphs-graphlet-functions/65/

szhorvat:
I added a few that came to mind, but without much organization. Feel free to remove/reorganize/rewrite as needed. In many cases, some functions are already exposed, but not with their full feature set.

cptwunderlich:
FYI, from a user perspective, I just wanted to check if a graph is chordal and typed it in, just to realize that `is_chordal` is not part of python-igraph. I think this would be a very useful function, not often found in other libraries.

szhorvat:
[...]

PR #391, opened 2021-04-29
Function parity with C core

iosonofabio:
First chunk of functions to be floated up to the Python interface from the C core, as planned in #389.

Some of this stuff in the C core is not yet released, so let's wait for the next minor release before merging this one.


szhorvat:
It would make sense to not do this as one huge PR, but break it down into several. I suggest renaming this PR to Eulerian cycles + Max cliques, making sure that these two work correctly, and merging.

iosonofabio:
👍

Issue #401, opened 2021-05-27
low efficiency of "from_networkx" compared to TupleList and the other way


jwang-19:
I encounter the slow efficiency problem of "from_networkx" when I was trying to use the Leiden algorithm on a very large network which is composed of around 33,000 nodes and 11.3 million edges. I reported the problem in leidenalg GitHub and the author Traag suggested me to report the issue here. So here are my experiments about the three ways to load the network:

"from_networkx":
```python
G1 = ig.Graph.from_networkx(G)
prepartition = la.find_partition(G1, la.RBConfigurationVertexPartition, None, G1.es["weight"], 20, 0, 1, resolution_parameter = inputResolution)
partition_dict = {}
for name, membership in zip(G1.vs["_nx_name"], prepartition.membership):
	partition_dict[int(name)] = int(membership)
```

"TupeList":
```python
G1 = ig.Graph.TupleList(edges, directed=False, vertex_name_attr="name", edge_attrs= ["weight", "estTime"], weights= False)
prepartition = la.find_partition(G1, la.RBConfigurationVertexPartition, None, G1.es["weight"], 20, 0, 1, resolution_parameter = inputResolution)
for name, membership in zip(G1.vs["name"], prepartition.membership):
	partition_dict[int(name)] = int(membership)
```

"the third method":
```python
G1 = ig.Graph(directed = False)
G1.add_vertices(list(set(G.nodes)))
G1.vs["name"] = list(set(G.nodes))
G1.add_edges([(x, y) for (x, y, z, w) in edges])
G1.es['weight'] = [z for (x, y, z, w) in edges]
G1.es['estTime'] = [w for (x, y, z, w) in edges]
prepartition = la.find_partition(G1, la.RBConfigurationVertexPartition, None, G1.es["weight"], 20, 0, 1, resolution_parameter = inputResolution)
for name, membership in zip(G1.vs["name"], prepartition.membership):
	partition_dict[int(name)] = int(membership)
```

The "from_networkx", "TupleList", and "creating a graph by adding vertices and edges" took around 14.7 minutes, 0.82 seconds, and 0.14 seconds accordingly. Such a distinct difference may be stem from the "networkx".

Also, I found the "from_networkx" can load all my nodes no matter the nodes have edges or not, but "TupleList" can only load the nodes that have edges, and the third way can load all nodes in ascending order. 

BTW, I used the `python-igraph` 0.8.3 and I obtained it from your site.

szhorvat:
[...]

PR #402, opened 2021-05-27
improve from_networkx performance

szhorvat:
This attempts to address #401.  It only fixes `from_networkx`, not `from_graph_tool`. I need some feedback before proceeding. I don't really know much Python, and this was merely an exercise for me. Take a very careful look before merging.

### DIFF
--- a/src/igraph/__init__.py
+++ b/src/igraph/__init__.py
@@ -1920,6 +1920,7 @@ class Graph(GraphBase):
         @param g: networkx Graph or DiGraph
         """
         import networkx as nx
+        from collections import defaultdict
 
         # Graph attributes
         gattr = dict(g.graph)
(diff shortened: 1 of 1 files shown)


szhorvat:
Benchmark:

```
import random
random.seed(123)
g = nx.gnm_random_graph(1000,10000)
```

then in ipython,

```
%timeit ig.Graph.from_networkx(g)
```

The speedup is from ~2.0 seconds to ~12 milliseconds.

szhorvat:
I notice that the name `Graph.from_networx` seems inconsistent with other graph constructors, all of which are capitalized. Shouldn't this be corrected?

szhorvat:
Why is the test for this functionality in `src/igraph/test` while everything else is in `tests`? Is this test run when doing `python -m unittest` as in the instructions?

Originally I messed up and forgot to convert vertex names to IDs, yet tests did not seem to fail. Assuming I ran the tests correctly, the tests should be improved to catch this mistake.

ntamas:
> I notice that the name Graph.from_networx seems inconsistent with other graph constructors, all of which are capitalized. Shouldn't this be corrected?

The capitalization is a leftover from the early days of the development of the Python interface where I thought this would be a good idea to distinguish class methods from "normal" methods. Nowadays I don't use this convention in new code and prefer normal `underscore_case` instead. When we re-design the `python-igraph` API in the future such that we won't have a monolithic `Graph` class any more, this inconsistency will probably not be relevant any more. Until then, we can alias `from_networkx` with `From_Networkx` if we want to. Opinions?

> Why is the test for this functionality in `src/igraph/test` while everything else is in tests? 

Because I probably forgot to move them when I switched from the tests-within-the-package layout to the tests-outside-the-package layout [as advocated here](https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure) :) It should be in `tests` by all means.

ntamas:
Thanks a lot!

szhorvat:
> Until then, we can alias `from_networkx` with `From_Networkx` if we want to. Opinions?

No preference from me, just want to note that if this is done, it should be `From_NetworkX` and not `From_Networkx`.

iosonofabio:
Thank you. Let's not rename and break API and also go against Python conventions.

Issue #395, opened 2021-05-20
Matplotlib backend does not support multi-edges


szhorvat:
**Describe the bug**
The matplotlib backend plots multi-edges on top of each other.

**To reproduce**
```
import matplotlib.pyplot as plt

g = ig.Graph.Formula('0-1-2-3-2',simplify=False)

fig, ax = plt.subplots()
ax.set_aspect('equal', 'box')
ig.plot(
    g,
    target=ax,
    vertex_size=8, edge_alpha=0.3,
)

plt.show()
```

![image](https://user-images.githubusercontent.com/1212871/119015088-e13d7180-b998-11eb-9866-396e3125374f.png)

**Version information**
0.9.1

ntamas:
Note to ourselves: the Cairo backend uses the [`autocurve()`](https://igraph.org/python/doc/api/igraph.html#autocurve) function to assign a "curvature" to each edge in a multiple-edge bundle. We can probably make use of this in the Matplotlib backend as well. Since `autocurve()` is potentially time-consuming, we use it only for graphs with less than 10K edges.

iosonofabio:
I'm pretty sure I messed with autocurve in the mpl backend, at least for loops. Let me have a look

ntamas:
This is now fixed in 0.9.4 so closing.

PR #406, opened 2021-05-30
fixed autocurve mpl

cvanelteren:
Autocurve was not working as expected for matplotlib plotting. This PR fixes it.

### DIFF
--- a/src/igraph/drawing/graph.py
+++ b/src/igraph/drawing/graph.py
@@ -1180,8 +1180,13 @@ class MatplotlibGraphDrawer(AbstractGraphDrawer):
             if default is True:
                 default = 0.5
             default = float(default)
-            autocurve(graph, attribute="curved", default=default)
-            kwds["edge_curved"] = graph.es.get_attribute_values("curved")
+            ecurved = autocurve(graph, attribute=None, default=default)
+        elif "edge_curved" in kwds:
+            ecurved = kwds["edge_curved"]
+        elif "curved" in graph.edge_attributes():
+            ecurved = graph.es["curved"]
+        else:
+            ecurved = [0] * ne
 
         # Arrow style for directed and undirected graphs
         if graph.is_directed():
@@ -1195,7 +1200,6 @@ class MatplotlibGraphDrawer(AbstractGraphDrawer):
 
         # Edge coordinates and curvature
         nloops = [0 for x in range(ne)]
-        has_curved = "curved" in graph.es.attributes()
         arrows = []
         for ie, edge in enumerate(graph.es):
             src, tgt = edge.source, edge.target
@@ -1275,7 +1279,7 @@ class MatplotlibGraphDrawer(AbstractGraphDrawer):
                 )
 
             else:
-                curved = edge["curved"] if has_curved else False
+                curved = ecurved[ie]
                 if curved:
                     aux1 = (2 * x1 + x2) / 3.0 - curved * 0.5 * (y2 - y1), (
                         2 * y1 + y2
--- a/src/igraph/drawing/graph.py
+++ b/src/igraph/drawing/graph.py
@@ -1180,7 +1180,8 @@ class MatplotlibGraphDrawer(AbstractGraphDrawer):
             if default is True:
                 default = 0.5
             default = float(default)
-            kwds["edge_curved"] = autocurve(graph, attribute=None, default=default)
+            autocurve(graph, attribute="curved", default=default)
+            kwds["edge_curved"] = graph.es.get_attribute_values("curved")
 
         # Arrow style for directed and undirected graphs
         if graph.is_directed():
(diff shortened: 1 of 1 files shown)


src/igraph/drawing/graph.py
```diff
@@ -1180,7 +1180,8 @@ def callback_edge_offset(event):
             if default is True:
                 default = 0.5
             default = float(default)
-            kwds["edge_curved"] = autocurve(graph, attribute=None, default=default)
+            autocurve(graph, attribute="curved", default=default)
+            kwds["edge_curved"] = graph.es.get_attribute_values("curved")
```
cvanelteren, line 1184: Btw. Not sure if this kwds is used in the code. Autocurve returns none --> prior code pushed None in this key. Now it contained the actual values.

src/igraph/drawing/graph.py
```diff
@@ -1276,12 +1277,12 @@ def callback_edge_offset(event):
             else:
                 curved = edge["curved"] if has_curved else False
                 if curved:
-                    aux1 = (2 * x1 + x2) / 3.0 - edge.curved * 0.5 * (y2 - y1), (
```
cvanelteren, line 1279: "curved" is not a property of the edge; either edge['curved'] should have been used, or curved as it was assigned to a var.

cvanelteren:
Example:

image

szhorvat:
Related: #395

cvanelteren:
Minor notes:

Either the `autocurve` has to invoked prior to calling the plotting or both edge_curved as well as autocurve has to be set in order for the plotting to work. 

That is

```python
ig.plot(g, target = ax, layout = layout,
        # vertex_size = 8
        edge_color = 'lightgray',
        edge_curved = -1, #any arbitrary angle will work
        autocurve = 1,
        )

```

works and 

```python
autocurve(g, "curved", 1)
ig.plot(g, target = ax, layout = layout)
```
works

iosonofabio:
hehe thank you for the PR! That's what happens when trying to tie new code into an existing infra without CI testing ;-)

I've been writing most of the mpl code including this bug, but @ntamas designed the previous Cairo interface so he might know more about the intention behind arguments vs attributes. Personally, I'd be fine with anything as long as there is consistency between the Cairo and mpl (for the time being).

Because it seems likely that we will soon switch to mpl as the main plotting beast of `python-igraph`, we'll have more freedom to break API then.

cvanelteren:
@iosonofabio No worries! Not too familiar with Cairo, but matplotlib is a great toolbox! I could possibly be of assistance there in the future if needed. Do you think  this PR can be merged as is? Not sure how much the api will change with the new move.

ntamas:
@cvanelteren I'll review this PR today and will most likely merge it. There are plans for a new `python-igraph` release in a few days, possibly with a joint patch release of `igraph` itself; I'll strive for including this PR in the next `python-igraph` release.

src/igraph/drawing/graph.py
ntamas: `autocurve()` is supposed to return `None` if and only if `attribute` is not `None`, otherwise it should return the calculated curvatures, so I'm not 100% sure whether this should be fixed here or in `autocurve()` itself. I'll investigate a bit.

ntamas: `autocurve()` seems to return a proper list here so this part of the patch is not needed; the real problem is that `kwds["edge_curved"]` is not used later. This part was copied from the `DefaultGraphDrawer` where `kwds` _is_ used later so that's the part that needs fixing.

ntamas:
Okay, so I took a closer look at the current Matplotlib graph drawer implementation and compared it to the default graph drawer. The default graph drawer uses multiple data sources for most of the visual attributes of vertices / edges; for instance, when it comes to determining the vertex color, it uses the `vertex_color` kwarg by default, but then falls back to the `color` vertex attribute if `vertex_color` is not specified. One problem with the current Matplotlib implementation is that this fallback mechanism is missing. Unfortunately the implementation of this mechanism in `DefaultGraphDrawer` is way too convoluted (I implemented it using metaclasses back in the Python 2.x era when I grew very fond of metaclasses, but it's totally unneeded here), so there's no point in porting it forward to `MatplotlibGraphDrawer`, especially if we want to push `MatplotlibGraphDrawer` as the primary one in the mid-term. However, such a fallback mechanism needs to be implemented at some point in the future if we want to bring `MatplotlibGraphDrawer` to feature parity with `DefaultGraphDrawer`.

In the short term, I'll just make it so that the Matplotlib drawer respects the `edge_curved` keyword argument, falling back to the `curved` attribute of the edge if `edge_curved` is not present. Right now it doesn't respect the `edge_curved` keyword argument, while `autocurve` works by injecthing this kwarg on-the-fly (to avoid modifying the edge attributes of the graph itself while plotting), so fixing the `edge_curved` kwarg should fix this issue as well.

ntamas:
Thanks a lot!

iosonofabio:
Thanks @ntamas 

Agree with you the infra for Cairo seems a little heavy for modern-day plotting - it works great though, which is what matters most ;-)

That's the main reason I didn't push any more features into the matplotlib stuff, I was a little unsure what parts of that infra you want to retain and port versus abandon. Perhaps we could have a quick chat about this in our next meeting.

Issue #398, opened 2021-05-20
Matplotlib plotting backend is not documented


szhorvat:
**Describe the bug**

Matplotlib plotting backend is not documented. 

See https://igraph.org/python/doc/api/igraph.drawing.html#plot

**Version information**
0.9.1

iosonofabio:
Hmm, I've documented it somewhere, but perhaps not everywhere. I'll take a look

szhorvat:
It would be great to have _some_ documentation on this before the next python-igraph release.

We might also want to mark it as "experimental". This will communicate that it's not finished functionality and it may change. The idea is to have things documented, so people can use them and give feedback, but not yet commit to the API or behaviour.

iosonofabio:
Alright, I had already documented it in the online manual here:

https://igraph.org/python/doc/tutorial/visualisation.html#plotting-graphs-within-matplotlib-figures

but the docstring of the function wasn't changed. My bad, let me add the info there too.

When explaining this kind of stuff, tutorials are most useful. The current tutorial:

https://igraph.org/python/doc/tutorial/tutorial.html#drawing-a-graph-using-a-layout

uses Cairo only, would you like me to add a matplotlib part to it? also the Python syntax in that tutorial is very 90s, pre PEP8 ;-)

szhorvat:
Re tutorial: I went through the plotting parts of the tutorial the other day (since I have a lot to learn in this area!) and noticed that the plots included in the tutorial are not quite the same as the what the preceding code produces. It would be nice to fix this. I'll let you know a bit later what the mismatch was (I need to look again).

iosonofabio:
I've now opened a PR on this, as you can see, happy to fix the tutorial based on what you find out!

Thank you!

PR #407, opened 2021-05-31
A bit more docs for matplotlib plotting

iosonofabio:
Addressing #398 with:

- a mention to matplotlib in the `ig.plot` docstring
- two mentions in the tutorial
- PEP8 linting of the tutorial, since we are there.

Let me know if we need anything else urgently (before next release) and I'll be happy to add it last minute!

### DIFF
--- a/doc/source/tutorial.rst
+++ b/doc/source/tutorial.rst
@@ -435,7 +435,7 @@ degree or betweenness centrality. You can do that with the tools presented so fa
 some basic Python knowledge, but since it is a common task to select vertices and edges
 based on attributes or structural properties, |igraph| gives you an easier way to do that:
 
->>> g.vs.select(_degree = g.maxdegree())["name"]
+>>> g.vs.select(_degree=g.maxdegree())["name"]
 ["Alice", "Bob"]
 
 The syntax may seem a little bit awkward for the first sight, so let's try to interpret
@@ -741,7 +741,7 @@ For instance, we can plot our imaginary social network with the Kamada-Kawai
 layout algorithm as follows:
 
 >>> layout = g.layout("kk")
->>> plot(g, layout = layout)
+>>> plot(g, layout=layout)
 
 This should open an external image viewer showing a visual representation of the network,
 something like the one on the following figure (although the exact placement of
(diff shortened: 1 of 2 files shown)


doc/source/tutorial.rst
```diff
@@ -753,6 +753,13 @@ nodes may be different on your machine since the layout is not deterministic):
 
    Our social network with the Kamada-Kawai layout algorithm
 
+If you prefer to use `matplotlib`_ as a default plotting engine, create an axes and use
+the ``target`` argument:
+
+>>> import matplotlib.pyplot as plt
+>>> fig, ax = plt.subplots()
+>>> plot(g, layout=layout, target=ax)
```
szhorvat, line 761: It seems to me that this does not make `matplotlib` the _default_. It is necessary to explicitly specify the `target` argument, i.e. modify the default.

That brings up the question: what if I want `matplotlib` to be the default? Should this be possible? Does the current design even allow making it the default, give that `target` takes an already constructed axes object? I just want a plot to show up in my notebook, and I want it to be made with matplotlib.

Keep in mind that these comments are coming from a Python novice who has very minimal experience with matplotlib. Take it as such: the perspective of a novice (which I think is useful, but should not always be used as the basis of decisions).

src/igraph/drawing/__init__.py
```diff
@@ -423,6 +423,10 @@ def plot(obj, target=None, bbox=(0, 0, 600, 600), *args, **kwds):
     @param target: the target where the object should be plotted. It can be one
       of the following types:
 
+        - C{matplotib.axes.Axes} -- a matplotlib/pyplot axes in which the
+          graph will be plotted. Drawing, saving to file, etc. is mostly
```
szhorvat, line 427: As a novice reading this, my immediate question is: what does "mostly" mean here?

doc/source/tutorial.rst
szhorvat: (I'm also hijacking this a bit to try to learn a bit more about Python ;-) )

iosonofabio:
Fair enough, I improved the wording.

I's quite easy to let people swap the default, but requires a little care since it's a **default** change and could be set per session, config file, module-wide, and/or for each import of `igraph`.

We don't need it for this release, but we should keep it in mind if we want to push mpl for the next one.

szhorvat:
What would you recommendation be if I wanted to have matplotlib as default today? Should I write a wrapper function for `plot` and let the default value of the `target` argument be an axes object created on the fly?

Once again, my comments here are more about my own learning than making suggestions.

iosonofabio:
define your axes and use the function with that target. That's how I use matplotlib for everything, not just igraph. Nonetheless, most tutorials encourage you to let something called `pyplot` handle some black magic for you, but that's generally a bad idea inherited from matlab, alas...

doc/source/tutorial.rst
ntamas: Discussion continues below in the main comment section.

ntamas:
@szhorvat So, I think that if you want Matplotlib to be the default, the best course of action is to override `plot()` with something like this:

```python
from igraph import plot as plot_original

def plot(*args, **kwds):
    if "target" not in kwds:
        fig, ax = plt.subplots()
        kwds["target"] = ax
    return plot_original(*args, **kwds)
```

This basically injects a fresh Matplotlib axes object if you don't have a target.

Issue #409, opened 2021-06-01
Incompatibility community detection and matplotlib backend


svalkiers:
Hi,

When using python-igraph's matplotlib backend, I have experienced some interoperability issues with the `mark_groups` functionality of the `plot` command. When I create a `VertexClustering` object (see working example) and try to visualize its groups without additional markup, everything works fine and I get a clean output as shown in the figure below.

```python
# WORKING EXAMPLE

import igraph as ig
import matplotlib.pyplot as plt
import random

random.seed(42)

visual_style = {}

g = ig.Graph.Erdos_Renyi(20,0.2)
communities = g.community_multilevel()
graph_layout = g.layout("kk")

visual_style["vertex_size"] = 10
# visual_style["vertex_color"] = "red"
visual_style["layout"] = graph_layout

fig, ax = plt.subplots()

ig.plot(communities,
        mark_groups=True,
        target=ax,
        **visual_style)
```

![image](https://user-images.githubusercontent.com/56355136/120326726-5427e000-c2e9-11eb-8083-3c45e97ff4a7.png)

However, when I try to add vertex color to the `visual_style` dictionary (i.e. uncomment line 16) , I get the following error:

```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-77-bab2d710ea9e> in <module>
     19 fig, ax = plt.subplots()
     20 
---> 21 ig.plot(communities,
     22         mark_groups=True,
     23         target=ax,

~/miniconda3/envs/immuno/lib/python3.9/site-packages/igraph/drawing/__init__.py in plot(obj, target, bbox, *args, **kwds)
    480     if hasattr(plt, "Axes") and isinstance(target, plt.Axes):
    481         result = MatplotlibGraphDrawer(ax=target)
--> 482         result.draw(obj, *args, **kwds)
    483         return
    484 

~/miniconda3/envs/immuno/lib/python3.9/site-packages/igraph/drawing/graph.py in draw(self, graph, *args, **kwds)
   1079         # Mark groups
   1080         if "mark_groups" in kwds:
-> 1081             for idx, color in kwds["mark_groups"].items():
   1082                 points = [vcoord[i] for i in idx]
   1083                 vertices = np.asarray(convex_hull(points, coords=True))

AttributeError: 'bool' object has no attribute 'items'
```

When using the regular plotting route, you can change the vertex color while marking the communities with the shaded area. Is this also possible with the matplotlib backend in any sort of way?

[...]

PR #410, opened 2021-06-02
Matplotlib plotting and clustering color bugfix

iosonofabio:
Tries to address #409 

The user's requested code now works on my machine in both cases, and the logic of the code is slightly improved and better documented in the code.

At some point we should consider adding *some* infrastructure to `MatplotlibGraphDrawer` to avoid having a lot of logic in a single giant function.

### DIFF
--- a/src/igraph/drawing/graph.py
+++ b/src/igraph/drawing/graph.py
@@ -1024,52 +1024,6 @@ class MatplotlibGraphDrawer(AbstractGraphDrawer):
             clustering = graph
             graph = clustering.graph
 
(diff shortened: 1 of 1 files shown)


ntamas (approved):

ntamas:
> At some point we should consider adding some infrastructure to `MatplotlibGraphDrawer` to avoid having a lot of logic in a single giant function.

Agreed! The same applies to the default graph drawer as well, and maybe there's the possibility of re-using some code between the two, so I'll look into this at some point in the future.

ntamas:
Also, thanks a lot for the quick PR!

PR #418, opened 2021-06-22
Reimplemented Graph.DataFrame with improved perfomance

fwitter:
New implementation is ~500 times faster than the current one.

### DIFF
--- a/src/igraph/__init__.py
+++ b/src/igraph/__init__.py
@@ -3401,22 +3401,21 @@ class Graph(GraphBase):
         return result
 
     @classmethod
-    def DataFrame(cls, edges, directed=True, vertices=None, use_vids=True):
+    def DataFrame(cls, edges, directed=True, vertices=None, use_vids=False):
         """Generates a graph from one or two dataframes.
 
         @param edges: pandas DataFrame containing edges and metadata. The first
           two columns of this DataFrame contain the source and target vertices
-          for each edge. These indicate the vertex IDs as nonnegative integers
-          rather than vertex *names* unless `use_vids` is False. Further columns
-          may contain edge attributes.
+          for each edge. These indicate the vertex *names* rather than IDs
+          unless `use_vids` is True and these are non-negative integers. Further
+          columns may contain edge attributes.
         @param directed: bool setting whether the graph is directed
         @param vertices: None (default) or pandas DataFrame containing vertex
-          metadata. The DataFrame's index must contain the vertex IDs as a
-          sequence of intergers from `0` to `len(vertices) - 1`. If `use_vids`
-          is False, the first column must contain the unique vertex *names*.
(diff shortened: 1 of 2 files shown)


iosonofabio:
Great to see work in this direction. Can you please explain where the performance increase is coming from? Just making sure we're not losing generality

fwitter:
For motivation and details, please see #419 .

With my reimplementation, I tried to stick to the behavior of the current implementation. However, I propose some changes:
- Additional checks for negative vertex IDs when `use_vids=True`
- Additional checks for valid index of the vertex DataFrame containing vertex IDs when `use_vids=True`
- Setting `use_vids=True` as the default to align with the rest of the DataFrame API, such that one could use `g_clone = Graph.DataFrame(g.get_edge_dataframe(), g.is_directed(), g.get_vertex_dataframe())` to clone a graph `g`

Moreover, I extended the unit test for `Graph.DataFrame`.

iosonofabio:
Thank you again. Please explain where the performance is coming from... The issue doesn't say either

fwitter:
Sure. I replaced the parts of the code which I identified to have the biggest impact on runtime. These are:

1. I replaced the call to `np.setdiff1d` by using Pandas `DataFrame.isin` (line 3495)
2. I changed the nested for-loops for adding vertex attributes one by one into a single for-loop iterating over columns and transferring each attribute for all vertices in batch (lines 3500-3501)
3. I changed the nested for-loops for adding edge attributes one by one into converting the edges DataFrame to a dictionary and passing the attributes directly to `Graph.add_edges` (lines 3505-3506)
4. I replaced the creation of the edges list using Python's `zip` by using `DataFrame.itertuples` (line 3504)

iosonofabio:
Thank you. I'm suspecting pandas might throw weird exceptions is we do batch operations hoping the data frames are consistent, did you check for those cases?

fwitter:
By consistency, you mean that the vertex IDs / names must be consistent throughout both edge and vertex DataFrames? This is checked and an exception is raised in case consistency is not given. The behavior did not change compared to the current implementation.

iosonofabio:
Thanks, sounds good. @ntamas ?

ntamas:
I'm not that familiar with Pandas, but it looks good to me! I have a slight problem with changing the default value of `use_vids=...` to `True` from `False` as it is technically a breaking change so I could not release this in a patch release as is. Would it be okay to change the default back to `False` temporarily and change it to `True` only in the next minor release?

fwitter:
Yes, I am fine with that. I reverted the change of the default value for `use_vids` and will create a new PR for that change which can be merged before the next minor release.

ntamas:
Great, thanks! I'll wait for the CI checks to complete and then I'll squash-and-merge this.

ntamas:
Thanks a lot!

PR #425, opened 2021-07-01
Revamping (yes!) the plotting interface: matplotlib and plotly

iosonofabio:
While `matplotlib` is supported for plotting now, the actual draw function is a hack.

By comparison, the `Cairo` drawing infrastructure is quite extensive - perhaps too much so.

Here I'm trying to clone the good parts of the Cairo interface into a revamped matplotlib interface, to stop the hacking from proliferating. I generally keep the inheritance a little shallower than it used to be, for sake of readability.

There is one issue on the table: matplotlib usually groups objects of the same kind for efficiency reasons into "collections". For instance, ATM all vertices are a single "collection" under `ax.get_children`. That grouping, which is reminescent of how SVG files work, is honestly convenient and enables changing color/sizes of all vertices at once. Ideally I'll try to have one collection for the vertices and one for the edges (not sure you can have one for the labels?), but that might fail.

I'll keep you posted.

**edit**: for folks who are new to this thread: I've redesigned the matplotlib plotting interface to integrate properly with the rest of the library, including:

1. switch to preexisting Cairo functions to compute plotting options: colors, convex hulls, edge curvature, loops, vertex size, etc.
2. add a configuration parameter: `plotting.backend` to switch btw `cairo` and `matplotlib`
3. add an optional legend to clustering plots
4. adapt and expand documentation on matplotlib plotting
5. add unit tests for several kinds of plots, including an infrastructure to run image comparisons on CI (adapted from matplotlib)

NOTE: I could expand point 5. to include tests for Cairo. The testing infrastructure would rely on matplotlib anyway though.

The main entry point for plotting is now `ig.plot`. For Cairo, that creates a `CairoPlot` class as before (was called `Plot`, that name is still available for API compatibility). For matplotlib, that uses or creates a matplotlib `Axes`, just like any other library on the planet that relies on matplotlib. `Axes` are abstractions that fulfill the role of `CairoPlot` in a generic way.

There's currently a few `#FIXME` tags across the codebase for small things, mostly deprecation warnings. After that's fixed, it'd be great to have one or two of you testing this and then merging.


szhorvat:
Question (really, just a question):

What will be the performance effects of switching to matplotlib?  How much performance do we lose or gain?  Is plotting gigantic graphs (which are becoming more and more common) still possible, as e.g. on Tamás's website? http://sixdegrees.hu/last.fm/

ntamas:
@szhorvat That graph on my website was created with a custom plotting function, not with igraph's built-in plot subsystem; it would have been too large for that, plus it did not offer the opportunity to put images as backgrounds in the nodes. (If you zoom in close enough on the interactive version of the artist map, you can actually see the profile pictures of the artists in the nodes).

I think that the current plotting backend is not really performant either, and I agree with @iosonofabio in the sense that it is probably a bit overengineered (this was my first attempt at a plotting backend). Also, I was in love with metaclasses at that time and I tried to solve everything with metaclasses :) So, making the Matplotlib backend more structured while keeping only the good bits of the current plotting backend is a good direction to go in.

iosonofabio:
A quick update in case you are interested.

Turns out many objects across the library have a `__plot__` method that contains some Cairo code. I'm trying to move that code into the `drawing` folder, using appropriately named modules. Instead, the actual `__plot__` function is just a passthrough that decides the backend (Cairo/Matplotlib), in analogy with the global `igraph.plot` function. A little bit of care is required to avoid cyclic deps.

This design might or might not stay depending on everyone's opinion, but at least it'd be great to move all floating chunks of Cairo code into `drawing`, so I think it's worth anyway and if we don't like it we can discuss better designs in this PR.

ntamas:
Thanks for the update @iosonofabio ; I'm keeping an eye on this PR.

Yes, ideally all Cairo-specific stuff should be hidden behind the Cairo backend in `igraph.drawing`; we should not leak these out to the `__plot__` methods.

iosonofabio:
Alright, three things obviously missing and then it's ready for discussion:

1. reimplementing Dendrogram plots in matplotlib. Not a biggie, turns out I have committed into matplotlib's dendrogram plots myself years ago ;-)
2. transferring the radius smoothener for polygons
3. edge labels and vertex/edge label positioning

Will take another day or two I think.

iosonofabio:
Alright, everything is implemented except for label positioning for vertices and edges. The main reason is that I'm a little unsure how matplotlib handles calculations for text box sizes across its own backends.

@ntamas if you could take an initial look at the many changes and advise on improvements, that'd be great. I'll try to figure out the text box thing in the meanwhile.

This should close a bunch of open bugs because it aligns the matplotlib infra with the existing Cairo, i.e. anything Cairo does right or wrong, mpl just follows.

Thanks!

iosonofabio:
There's one more thing I could do here, but it'd be good to hear opinions on it.

ATM in the single `Configuration` instance there are a few plotting options. We could add a "backend" option that sets the defaults to either Cairo or matplotlib. That would let people default their machine to mpl if they wished to. The default config *value* would still be Cairo for now.

If we do this (I'd be in favour) then it'd be a good idea to rename all those `DefaultXXXDrawer` into `CairoXXXDrawer` for clarity, and `PolygonDrawer` into `CairoPolygonDrawer`. That would break API big time.

What do you think?

ntamas:
I'll go through the commits in this PR shortly, but a few responses first:

* Adding a `backend` option to `Configuration` is a great idea, I totally agree with it.
* Renaming the `DefaultXXXDrawer` and the `PolygonDrawer` classes are also okay for me. We can keep temporary aliases to them until 0.10.x if somehow this gets released in 0.9.x. (I mean, just set `DefaultGraphDrawer = CairoGraphDrawer` and `PolygonDrawer = CairoPolygonDrawer` and we are good).

ntamas:
Okay, it seems like a common pattern that you need is "if the context seems to imply that the user wants the Matplotlib backend, then construct a `SomethingDrawer` object for the Matplotlib backend, otherwise construct an object for the Cairo backend. Plus, you need to pass the `bbox` and maybe the `palette` to the Cairo drawer but not to the Matplotlib backend. This should be factored out somehow, especially if we also allow the user to select a drawing backend in the configuration.

I'm starting to think that maybe the `__plot__` functions should already be given information about the backend they must work with, and the top-level `plot()` function should be the one responsible for picking a plotting backend (if the user did not specify one in the configuration; otherwise it should just use whatever the user wants).

Also, I wonder whether the `bbox`  and `palette` options would make sense for the Matplotlib backend or not. The purpose of the `bbox` parameter would be to let the user plot something to an arbitrary part of the canvas. I wonder whether that's possible with Matplotlib or not, and if not, we should re-consider the `Plot` object as well because the `Plot` object is essentially a glorified list of `(object, bbox, palette)` triplets that should be drawn on the same surface. The `palette` might be a bit problematic as well if Matplotlib also has its own concept of a palette (which it probably does); the original idea was that you could use numbers for coloring vertices and the palette would map the numbers to the corresponding colors. How shall we harmonize this with Matplotlib?

iosonofabio:
Thanks @ntamas 

Re the `Plot` class: I could keep a single class. The questionable thing is that each method would start with:
```python
if self._backend == 'matplotlib':
   (function body)
else:
   (function body)
```
which is quite an antipattern I think. Slightly better would be to have pairs of methods:
```python
class Plot:
    def _redraw_matplotlib(self):
        ...
    def _redraw_cairo(self)
        ...
    def redraw(self):
        '''(docstring)'''
       return getattr(self, '_redraw_'+self._backend)()
```
Next version would be to have two classes, `CairoPlot` and `MatplotlibPlot`, each with its own secret methods, and have `Plot` inheriting from both and implementing the generic passthrough methods (e.g. `redraw`).

This infrastructure, as well as bounding boxes, can be all done by hand in matplotlib, but most people wouldn't. `plt.Axes` is basically an abstraction layer that lets people do that (think of an "inset panel") without bothering with canvas-level coordinates (e.g. pixels). There is an object called a `Transform` that basically translates data coordinates or axes coordinates into canvas coordinates and back. So folks would opt for a dedicated `Axes` (perhaps with transparent background) instead of manually implementing the transforms as done here.

That works especially well in interactive backends, where you can dynamically resize the figure with the mouse, because then the absolute pixel coordinates of the plotted objects change (I do that all the time): matplotlib registers a callback function into the interactive backend that updates the `Transform` based on the figure size and triggers a redraw. Reimplementing all of that seems inconvenient.

Alright, in conclusion let me have a look and see if I can come up with something that makes sense.

szhorvat:
@iosonofabio Maybe [this](https://stackoverflow.com/q/68324445/695132) is of interest to you? Does matplotlib have built-in legending which can easily be made use of with the new matplotlib backend, but not with cairo?

iosonofabio:
Thanks @szhorvat 

yes matplotlib has infra for legends. That could be an easy option for clusters, happy to implement it if you think it'd be useful.

szhorvat:
I only have minimal experience with matplotlib so I can't judge if this should be built-in to igraph, or if matplotlib support has already made it much easier to add legends than it was with cairo (if yes, you might write an answer on StackOverflow?)  Just wanted to bring it to your attention that there is demand for legends. You'll know better than me how to best satisfy that demand, or if anything needs to be done on igraph's side.

iosonofabio:
Thanks, if there's a demand I'll try to include that in this PR.

iosonofabio:
plot_with_legend

iosonofabio:
I now tried to restructure the `Plot` infra via composition instead of inheritance. @ntamas would be great to hear your thoughts on that - well, hear as in read I s'pose ;-)

I also injected backend info into the `__plot__` methods, it cleans up a bit I must say, thanks for the idea.

I also implemented the legend @szhorvat mentioned, only in matplotlib, and without documentation for now (see above picture)

iosonofabio:
hey @ntamas 

I thought about it and came to the conclusion that the abstraction offered by `Plot` in Cairo is similar to matplotlib `Axes`, so it doesn't make much sense in layering them on top of one another.

You mentioned that originally `ig.plot` was a shorthand for `ig.Plot`, which was the preferred way to plot. I propose to flip that:

- `ig.plot` to become the preferred way to plot
- `ig.CairoPlot` (deprecating `ig.Plot`) to be used only for Cairo and only if more than one plot on the same surface is required.

To plot multiple things with matplotlib, one can just use the builting `Axes` as for any other library that uses mpl. Users will know how to do it (e.g. see `seaborn`, `networkx`, etc.)

iosonofabio:
I have now added testing infrastructure for matplotlib plotting. It is a simplified adaptation of matplotlib's own testing infra, and we should credit them somehow if we keep this.

Basically what this does is create figures with mpl and then testing the resulting SVG files against preexisting ones to check that everything was plotted as planned. We could extend to PNG and whatnot, but that might significantly increase the size of the git repo... even mpl is careful about that.

ntamas:
Thanks @iosonofabio , I'll take a closer look at it during the weekend and early next week!

ntamas:
Started review; sorry for the delay, last week turned out to be busier than expected.

ntamas:
@iosonofabio Is there a particular version of Matplotlib that I should use for testing? Lots of the unit tests seem to be failing; the images look different and there are also these failures:

```
ERROR: test_mark_groups (tests.drawing.test_graph.GraphTestRunner)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/tamas/dev/igraph/python-igraph/.venv/lib/python3.9/site-packages/igraph/drawing/graph.py", line 1054, in draw
    mg_iter = iter(kwds["mark_groups"])
TypeError: 'bool' object is not iterable
```

and

```
ERROR: test_mark_groups_squares (tests.drawing.test_graph.GraphTestRunner)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/Cellar/python@3.9/3.9.6/Frameworks/Python.framework/Versions/3.9/lib/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/Users/tamas/dev/igraph/python-igraph/tests/drawing/utils.py", line 56, in wrapper
    func(*args, **kwargs)
  File "/Users/tamas/dev/igraph/python-igraph/tests/drawing/test_graph.py", line 68, in test_mark_groups_squares
    dot.radius *= 0.5
AttributeError: 'PathCollection' object has no attribute 'radius'
```

iosonofabio:
Thank you @ntamas! Testing figure diffs pixel by pixel is a little delicate as you have seen - matplotlib has the same issue AFAIK.

I started making a galley of demos in `doc/source/examples/` to try and emulate matplotlib's gallery, the idea being that each example results in a certain figure that we ship similar to e.g.:

https://matplotlib.org/devdocs/gallery/lines_bars_and_markers/bar_label_demo.html#sphx-glr-download-gallery-lines-bars-and-markers-bar-label-demo-py

(i.e. having a "full code demo")

Do you think that's useful?

ntamas:
Yes, that's definitely useful! If you are eager to work on this now, feel free to push to this PR, but if it's not super urgent and/or you can do it in a separate branch, that would probably be better so I don't have to go back to reviewing files that you changed while I was going through them.

doc/source/visualisation.rst
```diff
@@ -137,13 +137,27 @@ you might want to change the size and color of the vertices:
...
->>> dots.set_sizes([250] * g.vcount())
+>>> dot = ax.get_children()[0] # This is a Circle for the first vertex
+>>> dot.set_color('tomato')
+>>> dot.radius *= 2 # double the default radius
 
 That also helps as a workaround if you cannot figure out how to use the plotting options below: just use the defaults and
 then customize the appearance of your graph via standard `matplotlib`_ tools.
 
+.. note:: The order of `ax.get_children()` is the following: (i) patches for clustering hulls if requested;
+   (ii) patches for vertices; (iii) patches for edges: for undirected graphs, there's one patch per edge. For directed graphs, there's a *pair* of patches, associated with the arrow body and head, respectively.
```
ntamas, line 148: Is this something that we can guarantee or is it something that depends on the internals of Matplotlib and may change in the future without notice if the Matplotlib folks decide to change things?

iosonofabio:
Thank you!

I've added support for plotting the histograms in matplotlib, it was easy enough.

As of the signatures of `__plot__`, I agree we should get rid of those None-defaulted `bbox` and `palette`. `CairoPlot` is already providing basically the *dataclass* functionality, having `self._palette` and `self.bbox` available. But then it also stores each object's bbox and palette as a simple tuple and those become positional arguments for `__plot__`. Basically we could glorify those tuples into a class.

The only issue I have with that is that it creates yet another level of abstraction. The alternative I would prefer would be to inject `bbox` and `palette` into the `kwds` dictionary that gets passed to `__plot__` anyway. Something like:

```python
        for obj, bbox, palette, opacity, args, kwds in self._objects:
            if palette is None:
                palette = getattr(obj, "_default_palette", self._palette)
            plotter = getattr(obj, "__plot__", None)

            [...]

           kwds['palette'] = palette
           kwds['bbox'] = bbox
     
            plotter("cairo", ctx, *args, **kwds)

```

that would leave most things as they are but get rid of the awkward signature for `__plot__`. Obviously, we would need to add a few backend-dependent checks about palettes and bboxes in the various `__plot__` functions.

doc/source/visualisation.rst
ntamas: The order is fine for me as is, I just wanted to make sure that this is something that we can control and not something that Matplotlib imposes on us.

ntamas:
@iosonofabio I noticed that you extended the list of supported image viewers in Linux. This is great, but note that the entire open-an-image-viewer-automatically feature is planned to be deprecated in 0.10.0 (see the docstring of the top-level `plot()` function and also `CairoPlot.show()`). If you have strong objections, we can keep the functionality, but then we need to figure out how to ensure that the temporary file that the graph was plotted into is still on the disk when the image viewer is opened but gets removed as soon as it is not needed any more. Actually, not being able to solve this in a reliable way was the priamry reason why we decided to deprecate this functionality.

So, all in all, do you mind if I simply remove `CairoPlot.show()` and all related comments and code from your PR? The PR would get merged into 0.10.0 only anyway, and the function is already marked for removal in 0.10.0.

iosonofabio:
No problem, let's remove the image viewer stuff, good catch!

ntamas:
whoops, I might have accidentally tagged a Github user in the previous commit message...

iosonofabio:
I added a drawer factory in `igraph.drawing.__init__.py` - hopefully what you wanted. Now each object's `__plot__` function is very simple (we could use inheritance, except for docstring):

```python
        from igraph.drawing import DrawerDirectory

        drawer = DrawerDirectory.resolve(self, backend)(context)
        drawer.draw(self, **kwds)
```
and the whole dispatchment business is handled inside the `drawing` submodule, which is nice: no plotting code lying around anymore!

This makes it quite easy to extend to plotly and whatnot, at least in theory: create a folder in `src/igraph/drawing` with the backend name, implement the drawers, and add them to the drawer directory, bum - done :smirk: 

No rush, I just had an hour on my hands and thought whynot

iosonofabio:
I started adding functions to support plotly for at least graphs. Doesn't appear to be too different from mpl, easier if at all. Will take another commit or two to finish it off.

iosonofabio:
Ok, I wrote tests for plotly and they pass locally (just making sure there are no runtime errors for now, I'm not yet checking for accuracy).

I think if the mpl part is solid we can merge and say plotly support is "experimental" and wait for feedback from the community

iosonofabio:
I now added an "image" comparison test suite for plotly along the lines of the matplotlib one.

Since plotly is web first, it makes JSON-style python dictionaries and then sends them to its beasty JS engine to evaluate. So one could do two things:

1. compare the JSONs (that's what I'm doing here) at varying degrees of detail (we can ask the JS engine to fill in the blanks for default plotting options, but that becomes hugely detailed)
2. compare the resulting SVG/PNG after conversion, as we do in mpl. That relies on an external converter though, which seems unstable upstream (currently `kaleido`, might change though).

So for the time being I just use a terse variation of option 1. I think it's better than nothing FTTB and we can always change it in the future.

I should make a raster comparison for Cairo too.

ntamas:
So I'm finally back to this PR and I'm trying to ensure that the plotting tests are also executed in the CI environment.

It looks like we have two separate problems; I'll continue tracking these down tomorrow:

* `module 'plotly' has no attribute 'graph_objects'` - I have no idea why the CI environment shows this; this works perfectly fine on my machine, although the Plotly tests fail later due to subtle differences in the JSON files (and I don't know yet whether it's a genuine problem or we need to relax the JSON comparison a bit).
* `Don't know how to convert .svg files to png` for Matplotlib plots. Again, I don't know why this happens in the CI because these tests work fine on my machine (after I moved the baseline images to the correct folder).

iosonofabio:
Thank you!

Good to hear both work on your local install. I'll take a look later today and see if I can figure it out.

iosonofabio:
Ok.

Point 1: sounds like an import error. We use "fake modules" for Cairo, mpl and plotly, and therefore if the import fails (e.g. we haven't installed plotly on the CI runner) it becomes an "empty" object that has no properties, and of course it has no "graph_objects" property either.

Point 2 (conversion to PNG) is clear from the matplotlib codebase: they require inkscape (I guess both of us have it installed on our local machines). For conversion from PDF to PNG, they use ghostscript instead.

We have a few options:
1. install inkscape on the CI server (added testing dep)
2. use PNG instead of SVG baseline images (added repo weight)
3. skip mpl tests on CI (added risk of bugs)
4. install gs on the CI server and switch to PDF (similar issue as solution 1, I guess)

It seems to me that the general question here is: do we want to run plotting tests on the CI server, even though we are skipping the numpy/pandas ones? or are we ok running them locally only?

What do you think?

vtraag:
There is one CI run (the source package) that also runs tests with panadas and numpy. We can add the dependencies for matplotlib and plotly also there, I guess. Inkscape will be a very heavy load thought, so if we can avoid that, that would be nice.

iosonofabio:
Thanks Vincent, do you feel like ghostscript would be a lighter load for the source CI server?

ntamas:
> do we want to run plotting tests on the CI server, even though we are skipping the numpy/pandas ones?

I have recently added a `testing` extra to `setup.py` and I am trying to install most of the optional dependencies in CI, at least in `master` (I might not have merged that change to this branch yet). Pandas won't work in the CI unfortunately because they do not seem to have the right wheels for PyPI so the CI system tries to compile it from source, which takes a lot of time and then eventually fails. But I am in general favour of testing as much as possible in CI unless a particular dependency adds significant time to the build. So, I agree with Vincent -- Inkscape is probably too heavyweight, but ghostscript seems to be a good middle ground.

iosonofabio:
I changed the baseline images to PDF and mpl seems to be able to compare them with test results without a file conversion. Let's see if that works on CI or I'll try installing ghostscript there and check again.

iosonofabio:
I'm finally opting for PNG baseline images, it's only a few hundred kBs anyway. I'll try to debug the plotly import asap so we can push this through.

iosonofabio:
It's still failing on CI which is really weird, since it works well on my laptop.

To be fair, `plotly.graph_objects` is a pretty wild beast if you look at the code, so it warrants a little more inspection

ntamas:
Reference to Plotly issue for posterity: https://github.com/plotly/plotly.py/issues/3377

ntamas:
Plotly-related tests also fail on my machine due to minor differences in the expected and the observed coordinates of the graph in the JSON representation; the differences do not seem to indicate a genuine failure but they are still large enough to warrant looking into this a bit more.

For instance, these are the observed X and Y coordinates in the `graph_basic.json` plot (obtained with `jq '.data[] | { x: .x[0], y: .y[0] }' graph_basic.json`):

```json
{
  "x": 1.0062902332686186,
  "y": -0.007847896326059573
}
{
  "x": 0.2810762367481289,
  "y": 0.9762414170533863
}
{
  "x": -0.8789517668412217,
  "y": 0.5906222417711454
}
{
  "x": -0.8706745044406359,
  "y": -0.631792828646428
}
{
  "x": 0.29446912864607855,
  "y": -1.001667715242344
}
```

and these are the expected ones:

```json
{
  "x": 1.015318095035966,
  "y": 0.03435580194714975
}
{
  "x": 0.29010409851547664,
  "y": 1.0184451153265959
}
{
  "x": -0.8699239050738742,
  "y": 0.6328259400443561
}
{
  "x": -0.8616466426732888,
  "y": -0.5895891303732176
}
{
  "x": 0.30349699041342515,
  "y": -0.9594640169691343
}
```

I don't know what causes the difference; any ideas?

iosonofabio:
Alright, I'm getting painfully aware of the constraints here. Seems like matplotlib conversion PDF->PNG requires ghostscript 9.0. My laptop has 9.54, but of course cibuildwheel images have 8.70 so mpl refuses to acknowledge that installation. After all, gs 9.0 was released quite recently, just short of [12 years ago](https://artifex.com/news/artifex-releases-ghostscript-9.0)... :sob:

But turns out cibuildwheels has options to specify the manylinux image to use, and manylinux2014, although a youngling at 8 years of age, might have ghostscrpt 9.0 which was already going to kindergarden at the time. I'm checking if that avenue is available, otherwise I'd be in favour of just using PNGs and debugging why the linux i686 platform produces large RMS differences compared to the expected outputs.

iosonofabio:
Ok, it now works **except** for an _installation_ issue on Pypy 3.6/3.7 across OSes which to me looks unrelated to matplotlib. @ntamas sorry for the verbose PR: do you have any idea what the pypy thing could be about?

iosonofabio:
Ok, finally skipping matplotlib tests on Pypy since they require numpy and that doesn't seem to install well on some pypy versions.

iosonofabio:
I think this is ready for (squash and?) merge. I'll open a separate PR later on to improve docs and tutorials, esp. with the new backends.

ntamas:
@iosonofabio PyPy is mostly a mystery to me as well;  I have never used it seriously. I wouldn't be surprised if someone told me that NumPy and/or Matplotlib is not compatible with PyPy at all, so I think it's okay to ignore the failure.

ntamas:
Re-targeting to `develop` as these changes are too extensive to be included in the next patch release.

ntamas:
Okay, all tests passed even after rebasing; let me go through the PR once again tomorrow and then I'll merge.

iosonofabio:
Fantastic than you

ntamas (approved):

ntamas:
:tada: !

iosonofabio:
Yaaay, revamp successful!

PR #427, opened 2021-07-12
add doc for font-family

szcf-weiya:
It seems no documentation on the customization of font-family, although this feature has been implemented in https://github.com/igraph/python-igraph/pull/32/commits
Without such documentation, until I came across an old question posted in https://stackoverflow.com/questions/24231430/how-to-use-custom-typeface-for-vertex-labels-in-python-igraph, and then checked the related source code, I got how to specify the font family.


ntamas (approved):

ntamas:
Thanks a lot!

PR #429, opened 2021-08-08
Integer transition

iosonofabio:
This PR aims to bubble the integer transition happening in the C core, which has just been merged into their `develop` branch, up to the Python layer.

**Progress:**
- [x] early deletion of `long_t` traces across the code
- [x] get to compile against the `develop` branch without errors (warnings left)
- [x] link it against `develop` -- OMG
- [x] create a graph with no edges 
- [x] create a graph with edges 
- [x] vertex attributes
- [x] edge lists and attributes
- [x] pass `test_basic.py`
- [x] pass `test_decomposition.py` with so many community detection algos
- [x] pass all tests except for callback functions
- [x] look into compiler warnings beyond unit tests
- [x] fix callback functions
- [x] actually get it to work ;-)

**Open issues:**
- [x] `igraph_matrix_t` gets converted using a recast: `PyLongFromLong((long)MATRIX(m)(...))`... created a separate infra for `matrix_int_t` now
- [x] `igraph_dominator_tree` uses integer vectors now, but the test expects `nan`?
- [x] `numpy_to_contiguous_memoryview` relied on sizing up `c_double`, that breaks now
- [x] `g.isomorphic_vf2` takes a callback function which has some finnicky type casts
- [x] `igraphmodule_i_Graph_motifs_randesu_callback` has some signature issue that makes the compiler suspicious

**Surprises:**
- diameter an shortest path functions seem to have an additional argument now, used to output the *edge paths*. I adapted the code with null pointers for now, but we should probably add a Python kwarg to let the user choose what they're looking for. However, it seems unrelated with the current PR so probably good for a future PR.

Will keep you guys posted, @ntamas feel free to take over when you're back from holiday if you want, no pressure.

edit: this is such an eye opener! I got tangled for 2 days in the intricacies of compiler and linker options... OMG. I got it now and am making progress, man!

Just for fun, here's the first error one gets:

```python
In [3]: g = ig.Graph(3, [(0, 1)])
MemoryError                               Traceback (most recent call last)
<ipython-input-3-079d422b887f> in <module>
----> 1 g = ig.Graph(3, [(0, 1)])

~/university/PI/projects/python-igraph/build/lib.linux-x86_64-3.9/igraph/__init__.py in __init__(self, *args, **kwds)
    321             GraphBase.__init__(self, __ptr=ptr)
    322         else:
--> 323             GraphBase.__init__(self, nverts, edges, directed)
    324 
    325         # Set the graph attributes

MemoryError: Error at src/core/vector.c:453: cannot reserve space for vector. -- Out of memory
```

guess it's time to buy a new machine :stuck_out_tongue_winking_eye:


iosonofabio:
Ok, good enough to deserve an external eye:

- There's two issues with callback functions, one raises a compiler warning and the other fails a test.
- `numpy_to_contiguous_memoryview` needs a look by @ntamas 

Apart from that, I think we can merge master into this, fix any residuals that result from the merge, and, **in coordination with the igraph repo**, merge this PR into master.

ntamas:
Everything is fixed now except the `numpy_to_contiguous_memoryview` part. I will create a `develop` branch on `python-igraph` and retarget this to `develop` because I want to keep the opportunity of releasing a new patch release of `python-igraph` from `master` if there is a hotfix that needs to be pushed out in a hurry.

Tests might still be breaking; I only tested things on my machine. Let's see how it goes.

ntamas:
Also, things still need to be tidied up after merging this PR; in particular, there are probably lots of unnecessary casts to `long` now that `igraph_integer_t` is actually a `long` or a `long long`. Casting blindly to a `long` might have problems if `sizeof(igraph_integer_t) > sizeof(long)`, so in the future I will grep the source code for explicit casts to see if we can get rid of them and/or handle overflows more gracefully.

iosonofabio:
Fantastic, this is sailing really smoothly @ntamas ! Thanks!

ntamas:
Yay, now it seems to work! I'll merge and then it's time for cleaning up the casts in the `develop` branch.

Thanks a lot for your work!

iosonofabio:
AMAZING! Thanks

Issue #431, opened 2021-08-20
Add support to release aarch64 wheels


odidev:
Problem
---------
On aarch64, ‘pip install python-igraph’ builds the wheels from source code and then installs it. It requires the user to have a development environment installed on his system. Also, it takes more time to build the wheels than downloading and extracting the wheels from PyPI.

Resolution
-----------
On aarch64, ‘pip install python-igraph’ should download the wheels from PyPI.

@ntamas, Please let me know your interest in releasing aarch64 wheels. I can help with this.

ntamas:
Thanks a lot! I have no objections against merging this if the CI tests run successfully in #432.

vtraag:
As an alternative, until binary wheels for `aarch64` become available, you can install `aarch64` packages from the `conda-forge` channel in `conda`, i.e. `conda install -c conda-forge python-igraph`.

ntamas:
The PR has been merged so closing this.

odidev:
@ntamas, can you please let me know if when AArch64 wheels are targeted for release?

ntamas:
There was a recent bugfix in the C core regarding Gomory-Hu trees that I would like to get included in the Python interface as well. So, there will be a new release of the C core this week, in the next few days, and shortly after that a Python patch release will also follow, also in this week. At least that's the plan.

PR #432, opened 2021-08-20
Add aarch64 wheel build support

odidev:
Added aarch64 wheel build support.
Related to https://github.com/igraph/python-igraph/issues/431, @ntamas Could you please review this PR?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,6 +34,31 @@ jobs:
         with:
           path: ./wheelhouse/*.whl
 
+  build_aarch64_wheels:
+    name: Build wheels on Linux AArch64
+    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+          fetch-depth: 0
+
+      - name: Set up QEMU
+        id: qemu
+        uses: docker/setup-qemu-action@v1
+
+      - name: Build wheels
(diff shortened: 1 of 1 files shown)


ntamas:
Thanks for the PR! Seems good to me; I'll let the jobs run now and see whether it works. If it does, I'll be happy to merge this.

ntamas:
@odidev It looks like the aarch64 build is taking quite a long time in CI; do you happen to know whether this is normal?

odidev:
> @odidev It looks like the aarch64 build is taking quite a long time in CI; do you happen to know whether this is normal?

@ntamas Yes. Because of QEMU, the AArch64 build takes longer time.

ntamas:
Thanks @odidev! I just came back from holidays so I'll take a look at the PR soon.

I see that you recently force-pushed to this PR. Are you planning to migrate it to Azure or will it stay on Travis CI? For me Travis would have been great but I realized that they don't have an option to store the build artifacts along with the build (you can only give it an external AWS bucket to upload stuff to). So, if you stick to Travis, I'll happily merge it but the built wheels will be made available on PyPI only after a release (you won't be able to download any wheels built for non-tagged commits). I think Azure can keep the build artifacts available for a while so we can run the job on all commits and you can get wheels for any of the commits that are in `master`.

vtraag:
Sorry, I thought that Azure was being used to build (`python`)-`igraph` in `conda-forge` for `aarch64`, but they are using Drone. This builds `igraph` in 6 minutes, so I guess that is using native `aarch64`.

PR #439, opened 2021-10-01
fix: handle text wrapping if width=0

jboynyc:
Current behavior:

```python
g = ig.Graph.Erdos_Renyi(5, m=4)
ig.plot(g, wrap_labels=True, vertex_size=[5, 5, 5, 5, 0], vertex_label=list('abcde'))
# -> UnboundLocalError: local variable 'iterlines' referenced before assignment
```

This fix makes sure that `iterlines` is properly set even if a vertex has size 0.


ntamas (approved):

ntamas:
Thanks a lot!

PR #440, opened 2021-10-02
Move Python methods into separate modules

iosonofabio:
`src/igraph/__init__.py` is very long, which makes it difficult to read. This PR aims at splitting out a bunch of functions and methods into separate Python modules to ameliorate that.

Hopefully that should help potential contributors as well as ourselves.

One nice side effect is that if we want to transition to SWIG/Pybind11/?? we could do that "one module at a time".

Suggestions for more methods/functions to move out of the `__init__.py` file?


ntamas:
Looking good! Let's try to get this done quickly because sometimes there are changes in the original implementations in `master` and it becomes harder to sync `master` with the PR the more time it takes.

E.g., there are these two PRs that should be incorporated into this PR as well before merging:

* https://github.com/igraph/python-igraph/commit/2ea15a612cf0b03db9a4aa0d6f00f8746c2041eb
* https://github.com/igraph/python-igraph/commit/21b4d8013f545729e29925a621323e558e1fd1c3

Let me know when it's ready to merge and I'll make it a priority.

iosonofabio:
I merged those two PRs and pushed ahead to modularise as much as possible.

The result is satisfactory I think, with around 1,100 LOC in `__init__.py` and a decently organised codebase for new contributors and just users who are debugging.

Let's give other folks another 2 days to comment and then I'll mark ready for review: you can then take a look and merge into `develop`.

ntamas:
Bump -- is this now ready for review? :)

iosonofabio:
I got sidetracked. Yes, please take a look. I expect a few questions but we shouldn't delay

iosonofabio:
Ouch, I was expecting this to trigger an error in the build, otherwise I could have stopped early. I'll look into option 1 for now.

If we redesign the API then we have to adapt all the `_igraph` C code as well... Uff. But the good news is that we could do that and then start switching modules to pybind11 one at a time.

One way to do this would be to have several folders in `src` and each contains both an extension and a pure Python module, so they are... Modular? What do you think?

iosonofabio:
Okay, I made #451 to implement option 1 for now. Take a look when you have a minute :-)

PR #442, opened 2021-10-04
fix: recognize SVG format from file extension

jboynyc:
The list in `Graph._identify_format` omitted the .svg extension. I added it and alphabetized the list to make it easier to spot other omissions.


ntamas (approved):
Thanks!

PR #451, opened 2021-10-29
Fix docstrings for pydoctor

iosonofabio:
I added a script that uses a little introspection to swap in the `Graph` methods into the virtual environment used for the documentation build.

I also copied the pydoctor patch into this repo since that's needed to run `scripts/mkdoc.sh` anyway.

@ntamas can you please take a look if this fixes the issue with the modularized methods? It's not super pretty but perhaps a decent enough band-aid until we migrate to code generation.

### DIFF
--- a/scripts/mkdoc.sh
+++ b/scripts/mkdoc.sh
@@ -26,6 +26,9 @@ fi
 
 PWD=`pwd`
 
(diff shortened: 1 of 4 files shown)


iosonofabio:
Hehe the fact that you don't like this makes you a healthy developer :-)

PR #453, opened 2021-11-02
Convert five statements to the usage of augmented assignments

elfring:
[Augmented assignment statements became available with Python 2](https://docs.python.org/3/whatsnew/2.0.html#augmented-assignment "Augmented assignments").
:thought_balloon: Thus improve five source code places accordingly.


ntamas (approved):
Thanks!

PR #457, opened 2021-11-13
Changed get_label_position func to account for edge curvature

Sriram-Pattabiraman:
get_label_position now computes the bezier curve for the edge that would be drawn, then finds the parameter based midpoint by putting `.5` as the parameter and taking the output as the label position.

### DIFF
--- a/src/igraph/drawing/edge.py
+++ b/src/igraph/drawing/edge.py
@@ -20,15 +20,18 @@ from math import atan2, cos, pi, sin, sqrt
 cairo = find_cairo()
 
 def get_bezier_control_points_for_curved_edge(x1, y1, x2, y2, curvature):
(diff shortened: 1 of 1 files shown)


ntamas:
This looks good, thanks! One problem that I have with this PR is that now there are three places in `igraph/drawing/edge.py` where we need to derive the control points of a Bézier curve that goes from one vertex of a graph to another with a given curvature. Would it be possible to refactor this logic into a separate private function (say `def _get_bezier_control_points_for_curved_edge(x0, y0, x1, y1, curvature)`) and call this private function from all the three places so we don't duplicate code unnecessarily any more? (Look for occurrences of `edge.curved` to find the other two places).

ntamas:
Looks good, thanks! I'll merge this if the CI tests pass.

ntamas:
CI failure is unrelated so merging, thanks a lot!

PR #468, opened 2021-12-02
test_atlas: skip graphs

vtraag:
When building `python-igraph` for `conda-forge`, we repeatedly run into problems with graph 70 in `test_atlast.py`. This PR pushes the fix we introduced in https://github.com/conda-forge/python-igraph-feedstock/pull/54 also here so that it will propagate to the `conda-forge` in the next build.


ntamas:
Thanks!

Issue #470, opened 2021-12-09
Shortest path function that returns a "shortest path tree" object usable for quick path retreival


szhorvat:
**What is the feature or improvement you would like to see?**

The current shortest path functions will return a _list of paths_ from a source `s` to all other vertices.

This is a proposal for an additional, different function which also takes a source `s` as input, and will return a special object which can be used for the efficient computation of paths from `s` to any `t` on demand. The object would contain a shortest path tree rooted at `s`, in a suitable representation. The C layer also produces this information, e.g. in 

```
int igraph_get_shortest_paths_dijkstra(const igraph_t *graph,
                                       igraph_vector_ptr_t *vertices,
                                       igraph_vector_ptr_t *edges,
                                       igraph_integer_t from,
                                       igraph_vs_t to,
                                       const igraph_vector_t *weights,
                                       igraph_neimode_t mode,
                                       igraph_vector_long_t *predecessors,
                                       igraph_vector_long_t *inbound_edges)
```

the `predecessors` and `inbound_edges` output arguments encode this tree.

The object will have methods to:

 - Get the shortest path to any `t` vertex
 - Extract the tree in various formats (edge ID list or an actual graph)

**Use cases for the feature**

The proposed object is a concise representation of the shortest path structure from a given source. It takes `O(|V|)` storage only. The full lists of paths takes `O(|V| d)` storage where d is the average distance from `s`. 

This object may be used repeatedly to get the shortest paths to arbitrary other nodes, very efficiently. If a user need to compute shortest path from `s` to other nodes, but does not know _which_ other nodes to take yet, they are likely to run a full single-source shortest path algorithm separate for each target node. This is inefficient—why don't we make it possible to store the gist of the result from a single run and re-use it?

This function will be especially useful when working with large graphs, where both computation time and storage requirements are a concern.

In some applications, the shortest path tree itself may be of use. It is useful to have a function to provide this in igraph. The proposed function would achieve this with a concise API and added functionality.

This is in the spirit of the [Flexibility design principle](https://github.com/igraph/igraph/wiki/Design-principles-for-functions#3-flexibility) I recently proposed: don't just return the result in the most naïve form; instead enable users to do as many different things as possible with the result of the computation.

Issue #469, opened 2021-12-08
Complete Python implementation of usability benchmarks


szhorvat:
We have a set of usability benchmarks—a collection of small tasks that help us judge how convenient it is to use the library. This issue is to implement each such task using python-igraph.

https://github.com/igraph/igraph/wiki/Usability-benchmarks

This collection of implementation will also be useful as a "dictionary" that helps users familiar with another high-level interface of igraph (or even a different graph library) get up to speed with python-igraph.

_If you plan to work on this issue, choose one or a few exercises of your liking, and post a comment noting which ones you will implementing._

yanchenm:
Hi, currently working on a PR for the following examples:

- Clique percolation
- Community visualization
- Visualize cliques

szhorvat:
@yanchenm Thanks for offering to contribute!

We created a new repository to host the solutions. You can open a pull request (a separate one for each task) there. https://github.com/igraph/usability-benchmarks  This is still a new project, and we are still exploring the best way to move forward with it. Things may change rapidly.

I would suggest opening a draft PR before you are completely done with the task. Create a new directory within `solutions/python-igraph` for each task. This way, we can see how you approach the task, and we can learn something not only about what is possible with python-igraph, but how easy it is a newcomer to discover the best way to do it.

dkhl65:
I am working with yanchenm on this issue and I completed the visualize cliques task.

PR #483, opened 2021-12-17
Shortest path function that returns a "shortest path tree" object usa…

arya-pudota:
Meeting request for quick path retreival #470

Use cases for the feature

The proposed object is a concise representation of the shortest path structure from a given source. It takes O(|V|) storage only. The full lists of paths takes O(|V| d) storage where d is the average distance from s.

This object may be used repeatedly to get the shortest paths to arbitrary other nodes, very efficiently. If a user need to compute shortest path from s to other nodes, but does not know which other nodes to take yet, they are likely to run a full single-source shortest path algorithm separate for each target node. This is inefficient—why don't we make it possible to store the gist of the result from a single run and re-use it?

This function will be especially useful when working with large graphs, where both computation time and storage requirements are a concern.

In some applications, the shortest path tree itself may be of use. It is useful to have a function to provide this in igraph. The proposed function would achieve this with a concise API and added functionality.

This is in the spirit of the Flexibility design principle I recently proposed: don't just return the result in the most naïve/simple form; instead enable users to do as many different things as possible with the result of the computation.

### DIFF
--- a/src/igraph/dijkstra.py
+++ b/src/igraph/dijkstra.py
@@ -1,0 +1,43 @@
+import sys
+
+
+def closest_neighbor(graph, weights, from_vertex):
+	min_dist = sys.maxint
+	for v in range(len(graph.vs[from_vertex])):
+		dist = weights[from_vertex][graph.vs[from_vertex][v]]
+		if (dist < min_dist):
+			min_dist = dist
+	
+	return min_dist
+
+
+def igraph_shortest_path_dijkstra(graph, from_vertex, weights):
+	no_of_nodes = graph.vcount()
+	no_of_edges = graph.ecount()
+	
+	if (len(weights) != no_of_edges):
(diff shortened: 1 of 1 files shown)


vtraag:
Thanks for being willing to contribute @arya-pudota ! However, what was described in #470 was to get the information from the C layer and create an efficient representation for the results in Python. You seem to try to recreate a Dijkstra algorithm from scratch in Python, which is not what is described in #470. This also requires some knowledge of how to interface between the C layer and the Python interface, which is more advanced stuff. If you are interested in learning this, this is of course possible, but requires some effort.

In the meantime, if you are interested in some other issues which can be worked on using Python code only, perhaps #469 could be of interest to you? This involves showing how you can solve certain tasks using the Python `igraph` library.

I am going to close this pull request for now.

PR #484, opened 2021-12-17
Updated Shortest Path Visualisation

Gomango999:
I've added edge labels describing the edge weights, and fixed the path to correctly reflect the one found in `g.get_shortest_path()`.

Previous Image
shortest_path

Updated Image
shortest_path


ntamas:
Can you try using a slightly darker gray for the edges (i.e. `g.es["color"] = "#666"` or something like that; this will make the numbers stand out more even if they are plotted right on top of the edge. Otherwise it looks good to me!

szhorvat:
In some other systems, including Mathematica, it is possible to set a background for labels. Setting the background to white solves the problem that @ntamas is talking about above.

If we want to get really fancy, we can do stuff like this:

image

It looks like the line is interrupted by the label.

Is something like this feasible / useful for python-igraph? @iosonofabio

iosonofabio:
Yep, all possible, but we don't support it yet. Wouldn't take a whole lot of effort to start supporting it

iosonofabio:
I added this right now in d1977f0. To use it:

```python
ig.plot(
    ...
    edge_background='white',
    edge_align_label=True,
)

Example:

```python
import igraph as ig
fig, ax = plt.subplots()
g = ig.Graph.Ring(8, directed=True)
ig.plot(g, target=ax, edge_background='#E44C', edge_label=['e1', 'e2', 'e3'], edge_align_label=True)
```

igraph_align_label_example


```

Gomango999:
Thanks for the fast changes as always @iosonofabio, the diagram looks so much better now :)

shortest_path

szhorvat:
I only saw this now. Very nice, @iosonofabio !!

iosonofabio:
Looks good, I made minor changes, merging.

PR #460, opened 2021-11-19
Updated Tutorial

Gomango999:
Hi all!

I'm Kevin, working <strike>under</strike> with @iosonofabio, and tasked with writing new tutorials for python-igraph. The new tutorials are targetted towards more concise, bite-sized articles which each present self-contained ideas, with a greater focus on fast and usable code examples. As a first commit, I've added a "quick start" tutorial, which includes ~50 lines of code for an individual who is already familiar with both python and graphs, who want a quick and dirty way to get started with igraph.

Please don't merge this yet, this is still a placeholder and I plan to add more to the list of tutorials. I'll let you all know when I need a review.

PR #486, opened 2021-12-17
Visualisise Betweeness

Gomango999:
This PR implements a tutorial for visualising both vertex and edge betweenness centrality measures in a graph, as  suggested [here](https://github.com/igraph/python-igraph/pull/460#issuecomment-981126320).

betweenness


ntamas:
We have a handy function named `rescale` in `igraph.utils` -- you can give it a sequence of input values and a desired output range, and it will rescale everything in a linear manner such that the minimum and the maximum lines up with the range that you have given to it. This can be used for determining the widths and sizes.

Also, I was wondering whether we should use a more straightforward graph. [Krackhardt's kite graph](https://www.wikiwand.com/en/Krackhardt_kite_graph) is typically used in textbooks to illustrate differences between various centrality measures, so maybe we could use that here as well?

iosonofabio:
I'd like to show both: a textbook example and a larger, messy graph (for data science folks)

vtraag:
Personally I also like the more messy results.

Gomango999:
Thanks as always for the feedback and tips and tricks :) I've switched to `ig.rescale`, and also included an example for the Krackhardt graph. The results are below:

betweenness

iosonofabio:
I've improved the matplotlib with panel labels and a fixed figure size, apart from that it looks good, merging.

doc/source/tutorials/betweenness/betweenness.rst
```diff
@@ -11,7 +11,7 @@ Visualizing Betweenness
 .. _edge_betweenness: https://igraph.org/python/doc/api/igraph._igraph.GraphBase.html#edge_betweenness
 .. |edge_betweenness| replace:: :meth:`edge_betweenness`
 
-This example will demonstrate how to visualize both vertex and edge betweenness with a custom defined color palette. We use the methods |betweenness|_ and |edge_betweenness|_ respectively.
+This example will demonstrate how to visualize both vertex and edge betweenness with a custom defined color palette. We use the methods |betweenness|_ and |edge_betweenness|_ respectively, and demonstrate the effects on a standard `Krackhardt Kite <https://www.wikiwand.com/en/Krackhardt_kite_graph>`_ graph, as well as a `Barabási-Albert <https://en.wikipedia.org/wiki/Barab%C3%A1si%E2%80%93Albert_model>`_ random graph.
```
szhorvat, line 14: will demonstrate -> demonstrates

szhorvat:
I don't know how I feel about the arbitrary non-linear rescaling of betweenness just to "make it look good" ... Whether we want it or not, people do learn network science by using igraph, and I fear that this is teaching a bad practice. What do you think @vtraag ? Maybe do the rescaling for the sizes (which is common: we can scale either the vertex radius or vertex area with some cantrality), but not the colour?

@Gomango999 Can we have a colour bar added to this plot? One of the major benefits that matplotlib brings is easy colour bars. Whatever you do with the non-linear rescaling, the colour bar's labels should show true betweenness values, not rescaled ones.

szhorvat:
You did the rescaling because preferential attachment networks have a very very skewed betweenness distribution. I would recommend a Watts-Strogatz network in 2D, e.g. a 10 by 10 grid, with rewiring probability between 0.05-0.1. This also gives a nice looking network plot, but with a more "normal" betweenness distribution.

image

szhorvat:
Also, it bothers me that there is a variable called `edge_betweenness` that does not actually store edge betweenness, but a non-linearly rescaled quantity.

Gomango999:
Cheers :) You were right that using Barabási-Albert was giving me skewed results, which encouraged me to scale the colours in a weird way so the graph didn't end up looking like just two colours. All these suggestions are completely fair, so I'll implement them when I get the time.

Happy holidays!

PR #487, opened 2021-12-19
WIP: Tutorial gallery

iosonofabio:
igraph_gallery_screenshot

Code adapted from seaborn.

Good? No good? Comments? Suggestions?


szhorvat:
Of course it's good!

I'd love to keep the aspect ratio of the figures intact if that easy, but unfortunately I can't offer a fix right now.

szhorvat:
The maximum bipartite matching should have a much stronger contrast between thick and thin edges. I'll fix it later (unless someone else gets to it first).

ntamas:
Here are [multiple solutions](https://stackoverflow.com/a/12912224/156771) for the aspect ratio problem.

iosonofabio:
Alright, thanks. I was more wondering if we want to have a graphics-driven gallery at all or we prefer text.

It sounds like having thumbnails would be easier to browse than just a text list, so I'll spend a little more time on this. Don't worry about the exact figures, I'll standardize that one so it makes sense and looks good.

szhorvat:
Yes, I think it would make sense.

iosonofabio:
igraph_gallery_v2

This is coming along nicely, now with animated GIFs whenever there is one in the tutorial subfolder (currently only for the animated ring).

iosonofabio:
Shall I just merge this?

If you want to test it, pull into a local branch and then:

```bash
./scripts/mkdoc.sh -t
```

ntamas:
Let me take a quick look at this as there are conflicts with `develop` at the moment.

ntamas:
I don't know why Github said that there are conflicts; it seems like there aren't any. Let me try this and then I'll merge.

ntamas:
Thanks a lot, this is now merged!

PR #488, opened 2021-12-19
Connected Components + Cluster Tutorial

Gomango999:
A tutorial showing how to visualise a node clustering, and then also contract it down into it's representative nodes (cluster graph).

clusters


doc/source/tutorials/contracting_clusters/assets/clusters.py
```diff
@@ -0,0 +1,45 @@
...
+    vertex_size=g1.vs['vertex_size'],
+    vertex_color=list(map(int, ig.rescale(clusters.membership, (0, 200), clamp=True))),
+    edge_width=0.7
+)
+
+# Contract vertices, using mean value for coordinates, and using the max vertex_size as the representation
+def mean(list):
+    return sum(list) / len(list)
+
+g2 = clusters.cluster_graph(combine_vertices={'x':mean, 'y':mean, 'vertex_size':max})
```
ntamas, line 29: `combine_vertices` understands `"mean"` (as a string) and `"max"` (also as a string) so there is no need to define `mean()` on your own. On the other hand, it's good to show that the `combine_vertices=...` kwarg works with arbitrary Python functions. The implementation behind `"mean"` and `"max"` (as strings) is implemented in C. Other options are: `"sum"`, `"prod"`, `"min"`, `"random"`, `"first"`, `"last"`, `"median"` and `"concat"` (for strings).

ntamas:
Unfortunately `clusters()` is a misnomer from the C library; the correct graph theory term would be "connected components" as people often think about graph clustering algorithms (e.g., spinglass clustering, modularity clustering, spectral clustering etc) when hearing the word "cluster". Even I was thinking "oh hey, why Kevin is not using a connected graph with a proper clustered structure like `Graph.GRG(100, 0.2)` in this example?" when I first read the topic of the PR.

So, I was wondering whether it would be better to call this example as being about "connected components". Or we could show the result of `Graph.community_fastgreedy()` on `Graph.GRG(100, 0.2)` instead, in the same way you do it now for connected components?

Either way, I leave it up to @iosonofabio to decide in the end.

iosonofabio:
Let's call this tutorial "Connected components"

szhorvat:
Doing the same with a community structure would actually make more sense. There is no point in contracting weakly connected components, but there are real use cases for doing the same with communities (clusters).

`Graph.SBM` is one easy way to generate a graph with a clear community structure.

Gomango999:
Yep, I agree. In hindsight, the use of contraction on already weakly connected components doesn't seem like a particularly instructive use case. I'll separate this tutorial out into two:
- One for colouring connected components
- The second for showing how to compress down communities.

szhorvat:
Using Graph.SBM may not be the best idea because it is likely to produce a graph where each cluster is connected to each other one. 

I gave the Les Miserables network a try (without weights, with edge betweenness communities), and I got this:

image

That's nice.

iosonofabio:
This tutorial is useful and an example of "more is more". So I'd suggest we take the opportunity to write two pieces, one about contracting connected components (as above) and one about contracting communities of a globally connected graph. The latter is more useful in many practical contexts, as @szhorvat mentioned, however putting both situations one after the other really shows what the difference is, which I think is educational for our readers.

@Gomango999 would you like to give it a go at showing both? The challenge of the second case is that the contracted nodes (each representing a community) should still be connected to one another. Happy to chat on the phone if you prefer!

Gomango999:
> @Gomango999 would you like to give it a go at showing both? 

Yep, 100% makes more sense to do a contraction on clusters which leave some nodes still connected to each other at the end. Happy to have a go at it tomorrrow :)

szhorvat:
Do you need me to send you the Les Miserables or the Dolphin networks? Both are commonly used to demonstrate clustering. They are also commonly found online.

iosonofabio:
@Gomango999 would be great if we had square figures instead of rectangular ones. One option would be to have two figures, one normal and one contracted, with two separate PNG files. Rationale will be clear in a few days as I push the visual gallery: it has to do with reliable thumbnailing.

Gomango999:
> Do you need me to send you the Les Miserables or the Dolphin networks? Both are commonly used to demonstrate clustering. They are also commonly found online.

No need to send them, I've found them online: [Les Miserables network](https://networkrepository.com/lesmis.php) and [Dolphin network](https://networkrepository.com/soc-dolphins.php). Thanks for the advice, both of these data sets looks great :)

szhorvat:
OK, good. Hopefully they're correct.

Be wary of https://networkrepository.com/ It is an extremely sloppy website that contains some corrupted datafiles, and often removes critical metadata. I recommend avoiding this site whenever possible. I had some bad experiences.

I'd suggest getting them from here:

http://www-personal.umich.edu/~mejn/netdata/

This version is an the "gml" format which igraph can read directly.

Gomango999:
> Be wary of https://networkrepository.com/ It is an extremely sloppy website that contains some corrupted datafiles, and often removes critical metadata. 

Yep you were right, the download buttons on the site are completely unresponsive :/ . Anyways, I've finished the tutorial cluster graph tutorial using the Les Miserables network. Here's a sneak peak at the generated graphs:

communities
cluster_graph

PR #489, opened 2021-12-19
Articulation Points + Bridges Tutorials

Gomango999:
Tutorials on how to visualise articulation points and bridges.

articulation_points

bridge2


ntamas:
Really nice plots! I didn't know `python-igraph` is capable of this :)

doc/source/tutorials/articulation_points/articulation_points.rst
```diff
@@ -0,0 +1,47 @@
...
+
+    # Plot graph
+    fig, ax = plt.subplots()
+    ig.plot(
+        g, 
+        target=ax, 
+        vertex_size=0.3,
+        vertex_color="lightblue",
+        vertex_label=range(g.vcount()),
+        vertex_frame_color = ["red" if i in articulation_points else "black" for i in range(g.vcount())],
```
ntamas, line 33: There's room for a bit of simplification here, but it's only a matter of taste.

If you do this above:

```
articulation_points = g.vs[g.articulation_points()]
```

you get the articulation points as a vertex sequence (`VertexSeq` object). Now, you can either do this:

```
plot(..., vertex_frame_color=["red" if v in articulation_points for v in g.vs], ...
```

which is a bit nicer than iterating over `range(g.vcount())`, or you can even do something like this beforehand:

```
g.vs["frame_color"] = "black"
articulation_points["frame_color"] = "red"
```

which sets up a `frame_color` attribute with a default value of `"black"` for all vertices, then overrides it to be `"red"` for the articulation points.

Just a suggestion, feel free to ignore if it doesn't fit with the direction the tutorial is going.

ntamas: I can see that you used a similar solution in the `bridges` tutorial, so you may as well leave things as is here and show the other way in the `bridges` tutorial.

iosonofabio: I'm with @ntamas on this one, avoid range if possible for once

iosonofabio: haha, our msg just crossed... either way is fine for me then

Gomango999: Yep, I had originally wanted to show both ways, but I like the idea of the vertex sequences a bit better, so I've switched to that instead. 

@iosonofabio should I keep the `vertex_label=range(g.vcount())` line, or should I switch it out with `vertex_label=['0', '1', '2', '3', ... , '8']` similar with how it was done for `shortest_paths.rst`?

szhorvat:
The bridges visualization is very nice!

doc/source/tutorials/articulation_points/articulation_points.rst
```diff
@@ -0,0 +1,49 @@
...
+
+This example shows how to compute and visualize the `articulation points <https://en.wikipedia.org/wiki/Biconnected_component>`_ in a graph using |articulation_points|_. For an example on bridges instead, see :ref:`tutorials-bridges`.
+
+.. code-block:: python
+
+    import igraph as ig
+    import matplotlib.pyplot as plt
+
+    # Construct graph
+    g = ig.Graph(edges=[(0, 1), (1, 2), (2, 0), (2, 3), (3, 4), (4, 5),
```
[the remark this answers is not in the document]
iosonofabio: Good idea, I added that, merging now

Gomango999:
Thanks for getting to it first @iosonofabio! I'm very curious about how Graph.Formula works, since it's [undocumented](https://igraph.org/python/doc/api/igraph.Graph.html#Formula) in the Python API. Perhaps it's worth writing a tutorial about?

ntamas:
You can find the docs [here](https://igraph.org/python/doc/api/igraph.formula.html#construct_graph_from_formula). This should be at `Graph.Formula` as well but something broke during a refactoring and PyDoctor does not recognize any more that these two functions are actually the same.

Gomango999:
Thanks for the link, I can see how the articulation points graph is defined alot cleaner using this feature. I think this is prime space for a tutorial, so let me just add this to my never ending list of ideas haha.

Issue #490, opened 2021-12-20
spanning_tree docs


Yaminyam:
**Describe the bug**
![image](https://user-images.githubusercontent.com/31057849/146734500-d7ea9a60-03dc-4226-a669-54d042fca152.png)
The description of the function describes MST, but the function name points to ST, so I actually executed the function.

The graph as shown in the following picture was used to obtain a tree with the corresponding function, but in reality it seems to return ST rather than MST.

**To reproduce**
Steps or minimal example code to reproduce the problem.

![image](https://user-images.githubusercontent.com/31057849/146734833-2e1b5760-6361-498f-9c20-6c8875c06b12.png)
![image](https://user-images.githubusercontent.com/31057849/146734845-3414eee7-0667-4fcc-b8de-99341fb0df71.png)

If you are confident that the issue is not in the Python interface but in the
C core of igraph, please add it to the main [igraph repo](https://github.com/igraph/igraph)
instead.

If you are unsure, feel free to add your issue here - we will transfer it to
the main [igraph repo](https://github.com/igraph/igraph) if the root cause is
in the C core of igraph.

**Version information**
Which version of `python-igraph` are you using and where did you obtain it?

vtraag:
You have to pass in the `weights` explicitly to `spanning_tree`, see the [docs](https://igraph.org/python/api/latest/igraph.Graph.html#spanning_tree).

iosonofabio:
@vtraag I asked @Gomango999 to perhaps make a tutorial on MST without/with weights?

Yaminyam:
@vtraag thank you! I have checked the link you sent.
However, there is a weight attr in the graph, so why not use it and pass a new weight vector as an argument?
And if it is a function that returns mst, wouldn't it be better to name the function as minimum_spanning_tree rather than spanning_tree?

iosonofabio:
@Yaminyam

1. silent assumptions lead to bugs, copying the weights out is literally 1 LOC - and you're guarateed to know what you're doing. Perhaps in the future if igraph becomes more intuitive all round this kind of things might change
2. historical reasons: changing the API breaks people's existing code, and should be done only if the benefit is clear

szhorvat:
I think the name `spanning_tree()` is good, regardless of backwards compatibility consideration.

[...]

PR #495, opened 2022-01-05
Spanning Tree + MST Tutorials

Gomango999:
Added a tutorial for both spanning trees as well as minimum spanning trees, in response to #490 

spanning_trees

minimum_spanning_trees


doc/source/tutorials/minimum_spanning_trees/assets/minimum_spanning_trees.py
```diff
@@ -0,0 +1,35 @@
+import igraph as ig
+import matplotlib.pyplot as plt
+import numpy as np
+
+# Generate grid graph with random weights
+np.random.seed(0)
+
+g = ig.Graph.Lattice([5, 5], circular=False)
+g.es["weight"] = np.random.randint(1, 20, g.ecount()).tolist()
```
ntamas, line 9: I'm not sure whether it's worth introducing NumPy as a dependency to this test case just to generate a few random numbers. You can achieve the same thing with `[randint(1, 20) for _ in g.es]`.

doc/source/tutorials/spanning_trees/assets/spanning_trees.py
```diff
@@ -0,0 +1,38 @@
...
+import matplotlib.pyplot as plt
+import numpy as np
+
+g = ig.Graph.Lattice([6, 6], circular=False)
+
+# Optional: Rearrange the vertex ids to get a more interesting spanning tree
+layout = g.layout("grid")
+
+np.random.seed(0)
+permutation = np.random.permutation(36)
```
ntamas, line 11: In the vein of attempting to keep `numpy` out of the examples (not because I don't like `numpy` but because you need to know `numpy` as well to understand what's going on here):

```
from random import shuffle
[...]
permutation = list(range(g.vcount()))
shuffle(permutation)
```

This has the added bonus that you don't need to call `.tolist()` on the permutation.

ntamas:
These figures look great! I've added a few comments to the code but this is not meant as a full review, I'll leave that to @iosonofabio

doc/source/tutorials/minimum_spanning_trees/assets/minimum_spanning_trees.py
Gomango999: Yep, numpy is more convenient in general, though I agree that it isn't necessary in this case, and can be avoided. I've made the changes suggested.

doc/source/tutorials/spanning_trees/assets/spanning_trees.py
Gomango999: Done.

doc/source/tutorials/minimum_spanning_trees/assets/minimum_spanning_trees.py
iosonofabio: Actually, this was still here, so I fixed it now.

PR #497, opened 2022-01-08
Delaunay Triangulation Tutorial

Gomango999:
A tutorial to plot the [Delaunay triangulation](https://en.wikipedia.org/wiki/Delaunay_triangulation) given a set of points, using [SciPy's Delaunay function](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.Delaunay.html). Additionally, there is an extra section on how to draw in the triangles as well using matplotlib.

delaunay-triangulation

delaunay-triangulation2


doc/source/tutorials/delaunay-triangulation/assets/delaunay-triangulation.py
```diff
@@ -0,0 +1,30 @@
+import igraph as ig
+import matplotlib.pyplot as plt
+import random
+from scipy.spatial import Delaunay
+
+# Generate a random geometric graph
+random.seed(0)
+g = ig.Graph.GRG(30, 0)
```
ntamas, line 8: I think this line deserves an explanation in the comments -- this one basically creates 30 random points in the 2D unit square and creates a graph where the zero distance guarantees that there will be no edges between the graphs.

Wouldn't it be clearer if we created a 30x2 random array with `numpy.random()` and then created a graph later on with `g = Graph(30); g.vs["x"], g.vs["y"] = coords[:, 0], coords[:, 1]`?

Gomango999:
Great thanks @iosonofabio. With this tutorial merged, this means that I've hit my own personal goal of 20 tutorials contributed as an intern! I'll be sticking around to contribute here and there, as well as finishing off my work on C igraph, but thanks everyone for the support and feedback. It's been super invaluable :)).

PR #500, opened 2022-01-09
Typo Fixes in Tutorials

Gomango999:
Just some minor typos I found while looking through the tutorials.


szhorvat:
Just wanted to point out that the Gallery, as well as tutorial titles, sometimes use title case and sometimes sentence case. It would be good to be consistent.

PR #509, opened 2022-02-17
Improve Documentation for contract_vertices

chaosflaws:
As I spent about half an hour this morning trying to figure out how `contract_vertices` actually works, here's a PR to improve the documentation a bit ;-)

I hope I have understood the method correctly now and that I didn't make a mistake when updating the documentation.


chaosflaws:
Turns out I *haven't* understood it completely yet - it seems that contracted vertices are only removed if they are at the end of the graph?

ntamas:
Thanks for the PR and sorry for the late reply; I am on holiday now and I'm checking emails only occasionally.

Anyhow, the `contract_vertices()` method is an interface to the underlying low-level `igraph_contract_vertices()` function in the C core of igraph. An igraph graph always has vertex IDs ranging from zero to `g.vcount()-1`, i.e. the number of vertices minus one. `contract_vertices()` contracts vertices by assigning new vertex IDs to them; essentially, two or more vertices are contracted to a single vertex if and only if they are assigned the same vertex ID after contraction. So, if you have three vertices (with IDs 0, 1 and 2) and you supply a mapping of `[0, 1, 0]` to this function, it means that vertex 0 will have ID 0 after the contraction, vertex 1 will have ID 1 after the contraction, and vertex 2 will have ID 0 as well - essentially meaning that the _original_ vertices 0 and 2 are contracted into a _single_ vertex with ID 0. Since the new graph does not have any vertices with IDs larger than 1, the new graph will have two vertices only; you can think of it as vertices being "removed" from the original graph, but what happens behind the scenes is that a completely new graph is constructed from scratch, remapping the vertex IDs in the edge list of the original graph to the new vertex IDs.

Now, if the mapping that you provide to the function in the above example is not `[0, 1, 0]` but `[0, 2, 0]`, this means that the original vertices 0 and 2 will be remapped to the new vertex ID 0, and the original vertex 1 will be remapped to the new vertex ID 2. Since vertex IDs in igraph graphs always have to cover an integer interval, and the largest vertex ID is 2, it follows that the new graph must have three vertices, so nothing is "removed" in this case.

Does this help in understanding how the contraction works? If so, can you update the PR so that it makes things clearer to you, given the additional info I provided here?

@szhorvat One thing I was wondering is whether it would be a good idea to add a `min_vertex_count` argument to the underlying `igraph_contract_vertices()` function in the C core, which would allow the user to keep at least as many vertices as there are in the original graph, even if the mapping does not use all the vertex IDs available in this range? Note that in this case, the operation is not really a "contraction" but more like a genuine remapping of vertex IDs.

szhorvat:
> @szhorvat One thing I was wondering is whether it would be a good idea to add a `min_vertex_count` argument to the underlying `igraph_contract_vertices()` function in the C core, which would allow the user to keep at least as many vertices as there are in the original graph

I don't see any good use cases for that, or for creating a graph which has more vertices than the number of groups (even though that is technically possible by using higher vertex IDs than the number of groups).

What would be more useful on the Python side is to make it possible to input sets of vertices that need to be contracted. I might just want to contract two vertices and leave the rest alone. Manually creating the mapping for this is now tedious.

PR #511, opened 2022-02-21
Vector pointers -> Typed vector lists

iosonofabio:
This PR aims at broadcasting to Python the recent introduction of typed vector lists in the igraph C core.

### DIFF
--- a/src/_igraph/graphobject.c
+++ b/src/_igraph/graphobject.c
@@ -11082,38 +11082,38 @@ PyObject *igraphmodule_Graph_minimum_size_separators(
 PyObject *igraphmodule_Graph_cohesive_blocks(igraphmodule_GraphObject *self,
     PyObject *args, PyObject *kwds) {
   PyObject *blocks_o, *cohesion_o, *parents_o, *result_o;
-  igraph_vector_ptr_t blocks;
+  igraph_vector_int_list_t blocks;
   igraph_vector_int_t cohesion, parents;
 
-  if (igraph_vector_ptr_init(&blocks, 0)) {
+  if (igraph_vector_int_list_init(&blocks, 0)) {
     igraphmodule_handle_igraph_error();
     return NULL;
   }
 
   if (igraph_vector_int_init(&cohesion, 0)) {
-    igraph_vector_ptr_destroy(&blocks);
+    igraph_vector_int_list_destroy(&blocks);
     igraphmodule_handle_igraph_error();
     return NULL;
   }
 
   if (igraph_vector_int_init(&parents, 0)) {
-    igraph_vector_ptr_destroy(&blocks);
+    igraph_vector_int_list_destroy(&blocks);
     igraph_vector_int_destroy(&cohesion);
     igraphmodule_handle_igraph_error();
     return NULL;
   }
 
   if (igraph_cohesive_blocks(&self->g, &blocks, &cohesion, &parents, 0)) {
-    igraph_vector_ptr_destroy(&blocks);
+    igraph_vector_int_list_destroy(&blocks);
     igraph_vector_int_destroy(&cohesion);
     igraph_vector_int_destroy(&parents);
     igraphmodule_handle_igraph_error();
     return NULL;
   }
 
-  blocks_o = igraphmodule_vector_int_ptr_t_to_PyList(&blocks);
-  IGRAPH_VECTOR_PTR_SET_ITEM_DESTRUCTOR(&blocks, igraph_vector_int_destroy);
-  igraph_vector_ptr_destroy_all(&blocks);
+  blocks_o = igraphmodule_vector_int_list_t_to_PyList(&blocks);
+  IGRAPH_VECTOR_INT_LIST_SET_ITEM_DESTRUCTOR(&blocks, igraph_vector_int_destroy);
+  igraph_vector_int_list_destroy(&blocks);
   if (blocks_o == NULL) {
     igraph_vector_int_destroy(&parents);
     igraph_vector_int_destroy(&cohesion);
@@ -11160,8 +11160,8 @@ PyObject *igraphmodule_Graph_cliques(igraphmodule_GraphObject * self,
   static char *kwlist[] = { "min", "max", NULL };
   PyObject *list, *item;
   Py_ssize_t min_size = 0, max_size = 0;
-  igraph_integer_t i, j, n;
-  igraph_vector_ptr_t result;
+  igraph_integer_t i, n;
+  igraph_vector_int_list_t result;
 
   if (!PyArg_ParseTupleAndKeywords(args, kwds, "|nn", kwlist,
                                    &min_size, &max_size))
@@ -11179,37 +11179,34 @@ PyObject *igraphmodule_Graph_cliques(igraphmodule_GraphObject * self,
     max_size = -1;
   }
 
-  if (igraph_vector_ptr_init(&result, 0)) {
+  if (igraph_vector_int_list_init(&result, 0)) {
     PyErr_SetString(PyExc_MemoryError, "");
     return NULL;
   }
 
   if (igraph_cliques(&self->g, &result, min_size, max_size)) {
-    igraph_vector_ptr_destroy(&result);
+    igraph_vector_int_list_destroy(&result);
     return igraphmodule_handle_igraph_error();
   }
 
-  n = igraph_vector_ptr_size(&result);
+  n = igraph_vector_int_list_size(&result);
   list = PyList_New(n);
   if (!list) {
     return NULL;
   }
 
   for (i = 0; i < n; i++) {
-    igraph_vector_int_t *vec = (igraph_vector_int_t *) VECTOR(result)[i];
+    igraph_vector_int_t *vec = igraph_vector_int_list_get_ptr(&result, i);
     item = igraphmodule_vector_int_t_to_PyTuple(vec);
     if (!item || PyList_SetItem(list, i, item)) {
-      for (j = i; j < n; j++) {
-        igraph_vector_int_destroy((igraph_vector_int_t *) VECTOR(result)[j]);
-      }
-      igraph_vector_ptr_destroy_all(&result);
+      igraph_vector_int_list_destroy(&result);
       Py_XDECREF(item);
       Py_DECREF(list);
       return NULL;
     }
     igraph_vector_int_destroy(vec);
   }
-  igraph_vector_ptr_destroy_all(&result);
+  igraph_vector_int_list_destroy(&result);
 
   return list;
 }
@@ -11220,40 +11217,37 @@ PyObject *igraphmodule_Graph_cliques(igraphmodule_GraphObject * self,
 PyObject *igraphmodule_Graph_largest_cliques(igraphmodule_GraphObject * self)
 {
   PyObject *list, *item;
-  igraph_integer_t i, j, n;
-  igraph_vector_ptr_t result;
+  igraph_integer_t i, n;
+  igraph_vector_int_list_t result;
 
-  if (igraph_vector_ptr_init(&result, 0)) {
+  if (igraph_vector_int_list_init(&result, 0)) {
     PyErr_SetString(PyExc_MemoryError, "");
     return NULL;
   }
 
   if (igraph_largest_cliques(&self->g, &result)) {
-    igraph_vector_ptr_destroy(&result);
+    igraph_vector_int_list_destroy(&result);
     return igraphmodule_handle_igraph_error();
   }
 
-  n = igraph_vector_ptr_size(&result);
+  n = igraph_vector_int_list_size(&result);
   list = PyList_New(n);
   if (!list) {
     return NULL;
   }
 
   for (i = 0; i < n; i++) {
-    igraph_vector_int_t *vec = (igraph_vector_int_t *) VECTOR(result)[i];
+    igraph_vector_int_t *vec = igraph_vector_int_list_get_ptr(&result, i);
     item = igraphmodule_vector_int_t_to_PyTuple(vec);
     if (!item || PyList_SetItem(list, i, item)) {
-      for (j = i; j < n; j++) {
-        igraph_vector_int_destroy((igraph_vector_int_t *) VECTOR(result)[j]);
-      }
-      igraph_vector_ptr_destroy_all(&result);
+      igraph_vector_int_list_destroy(&result);
       Py_XDECREF(item);
       Py_DECREF(list);
       return NULL;
     }
     igraph_vector_int_destroy(vec);
   }
-  igraph_vector_ptr_destroy_all(&result);
+  igraph_vector_int_list_destroy(&result);
 
   return list;
 }
@@ -11316,8 +11310,8 @@ PyObject *igraphmodule_Graph_maximal_cliques(igraphmodule_GraphObject * self,
     PyObject* args, PyObject* kwds) {
   static char* kwlist[] = { "min", "max", "file", NULL };
   PyObject *list, *item, *file = Py_None;
-  Py_ssize_t min = 0, max = 0, i, j, n;
-  igraph_vector_ptr_t result;
+  Py_ssize_t min = 0, max = 0, i, n;
+  igraph_vector_int_list_t result;
   igraphmodule_filehandle_t filehandle;
 
   if (!PyArg_ParseTupleAndKeywords(args, kwds, "|nnO", kwlist, &min, &max, &file))
@@ -11327,37 +11321,34 @@ PyObject *igraphmodule_Graph_maximal_cliques(igraphmodule_GraphObject * self,
   CHECK_SSIZE_T_RANGE(max, "maximum size");
 
   if (file == Py_None) {
-    if (igraph_vector_ptr_init(&result, 0)) {
+    if (igraph_vector_int_list_init(&result, 0)) {
       PyErr_SetString(PyExc_MemoryError, "");
       return NULL;
     }
 
     if (igraph_maximal_cliques(&self->g, &result, min, max)) {
-      igraph_vector_ptr_destroy(&result);
+      igraph_vector_int_list_destroy(&result);
       return igraphmodule_handle_igraph_error();
     }
 
-    n = igraph_vector_ptr_size(&result);
+    n = igraph_vector_int_list_size(&result);
     list = PyList_New(n);
     if (!list) {
       return NULL;
     }
 
     for (i = 0; i < n; i++) {
-      igraph_vector_int_t *vec = (igraph_vector_int_t *) VECTOR(result)[i];
+      igraph_vector_int_t *vec = igraph_vector_int_list_get_ptr(&result, i);
       item = igraphmodule_vector_int_t_to_PyTuple(vec);
       if (!item || PyList_SetItem(list, i, item)) {
-        for (j = i; j < n; j++) {
-          igraph_vector_int_destroy((igraph_vector_int_t *) VECTOR(result)[j]);
-        }
-        igraph_vector_ptr_destroy_all(&result);
+        igraph_vector_int_list_destroy(&result);
         Py_XDECREF(item);
         Py_DECREF(list);
         return NULL;
       }
       igraph_vector_int_destroy(vec);
     }
-    igraph_vector_ptr_destroy_all(&result);
+    igraph_vector_int_list_destroy(&result);
 
     return list;
   } else {
@@ -11398,8 +11389,8 @@ PyObject *igraphmodule_Graph_independent_vertex_sets(igraphmodule_GraphObject
   static char *kwlist[] = { "min", "max", NULL };
   PyObject *list, *item;
   Py_ssize_t min_size = 0, max_size = 0;
-  igraph_integer_t i, j, n;
-  igraph_vector_ptr_t result;
+  igraph_integer_t i, n;
+  igraph_vector_int_list_t result;
 
   if (!PyArg_ParseTupleAndKeywords(args, kwds, "|nn", kwlist,
                                    &min_size, &max_size))
@@ -11417,37 +11408,34 @@ PyObject *igraphmodule_Graph_independent_vertex_sets(igraphmodule_GraphObject
     max_size = -1;
   }
 
-  if (igraph_vector_ptr_init(&result, 0)) {
+  if (igraph_vector_int_list_init(&result, 0)) {
     PyErr_SetString(PyExc_MemoryError, "");
     return NULL;
   }
 
   if (igraph_independent_vertex_sets(&self->g, &result, min_size, max_size)) {
-    igraph_vector_ptr_destroy(&result);
+    igraph_vector_int_list_destroy(&result);
     return igraphmodule_handle_igraph_error();
   }
 
-  n = igraph_vector_ptr_size(&result);
+  n = igraph_vector_int_list_size(&result);
   list = PyList_New(n);
   if (!list) {
     return NULL;
   }
 
   for (i = 0; i < n; i++) {
-    igraph_vector_int_t *vec = (igraph_vector_int_t *) VECTOR(result)[i];
+    igraph_vector_int_t *vec = igraph_vector_int_list_get_ptr(&result, i);
     item = igraphmodule_vector_int_t_to_PyTuple(vec);
     if (!item || PyList_SetItem(list, i, item)) {
-      for (j = i; j < n; j++) {
-        igraph_vector_int_destroy((igraph_vector_int_t *) VECTOR(result)[j]);
-      }
-      igraph_vector_ptr_destroy_all(&result);
+      igraph_vector_int_list_destroy(&result);
       Py_XDECREF(item);
       Py_DECREF(list);
       return NULL;
     }
     igraph_vector_int_destroy(vec);
   }
-  igraph_vector_ptr_destroy_all(&result);
+  igraph_vector_int_list_destroy(&result);
 
   return list;
 }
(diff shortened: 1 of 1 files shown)


src/_igraph/convert.c
```diff
@@ -2684,6 +2684,73 @@ int igraphmodule_PyObject_to_vector_int_ptr_t(PyObject* list, igraph_vector_ptr_
...
+    igraphmodule_handle_igraph_error();
+    Py_DECREF(it);
+    return 1;
+  }
+
+  /* TODO: what about this? is memory managed already? */
+  //IGRAPH_VECTOR_PTR_SET_ITEM_DESTRUCTOR(vec, igraph_vector_int_destroy);
+  while ((item = PyIter_Next(it)) != 0) {
+    /* FIXME: is this even the right way to do this? */
+    subvec = igraph_Calloc(1, igraph_vector_t);
```
szhorvat, line 2720: igraph_Calloc is deprecated, use IGRAPH_CALLOC

src/_igraph/convert.c
```diff
@@ -2684,6 +2684,73 @@ int igraphmodule_PyObject_to_vector_int_ptr_t(PyObject* list, igraph_vector_ptr_
...
+/**
+ * \ingroup python_interface_conversion
+ * \brief Converts a Python list of lists to an \c igraph_vector_list_t
+ *        (containing \c igraph_vector_int_t items).
+ *
+ * \param o the Python object representing the list of lists
+ * \param m the address of an uninitialized \c igraph_vector_list_t
+ * \return 0 if everything was OK, 1 otherwise. Sets appropriate exceptions.
+ */
+int igraphmodule_PyObject_to_vector_list_t(PyObject* list, igraph_vector_list_t* vec) {
```
szhorvat, line 2696: It's much better to call `vector_list_t` objects a "list" so they are never confused with vectors. `list` is a better name than `vec`.

src/_igraph/convert.c
```diff
@@ -2684,6 +2684,73 @@ int igraphmodule_PyObject_to_vector_int_ptr_t(PyObject* list, igraph_vector_ptr_
...
+  if (igraph_vector_list_init(vec, 0)) {
+    igraphmodule_handle_igraph_error();
+    Py_DECREF(it);
+    return 1;
+  }
+
+  /* TODO: what about this? is memory managed already? */
+  //IGRAPH_VECTOR_PTR_SET_ITEM_DESTRUCTOR(vec, igraph_vector_int_destroy);
+  while ((item = PyIter_Next(it)) != 0) {
+    /* FIXME: is this even the right way to do this? */
```
szhorvat, line 2719: No, this is not right.

You are using `push_back_new()` below. That function creates a new vector from scratch, adds it to the list, and returns a pointer to it. You cannot pass in your own vector. 

What you probably want is `push_back()`, which transfers the ownership of the vector. This means that you should _not_ destroy `subvec` once it's been inserted into the list. You should not touch it at all—it's gone, it lives in the list now. To access it, index into the list.

You do not need to allocate `subvec` on the heap. Keep things simple, use a local variable.

(But do not name these vec and subvec if possible, name them list and vec)

iosonofabio: yes, transfer ownership is what we need here, thank you for the hint

Just wording, but remember I didn't "name" any of those variables, the names were already there. I agree we should try and change names to something less confusing as we convert the whole codebase. Same for "keeping things simple":  I'm not coming up with these algorithms, they are already there, so I cannot "keep" things simple, but only *simplify* compared to what it is now. In any case, happy to simplify whenever possible, of course.

szhorvat: Sorry, I didn't see the preceding function that you were basing this on. Yes, that was a "pointer vector" containing "vectors" so "vec of vecs" was consistent naming.  It also necessitated first allocating a vector_t dynamically before inserting it into the pointer vector.  Now we don't have to do that anymore.

Let me know if you have questions about `vector_list_t` until Tamás comes back. I don't know the Python API but I do know `vector_list_t` and how to move from a `vector_ptr_t`.

src/_igraph/convert.c
```diff
@@ -2617,6 +2684,140 @@ int igraphmodule_PyObject_to_vector_int_ptr_t(PyObject* list, igraph_vector_ptr_
...
+  }
+
+  if (igraph_vector_list_init(vec, 0)) {
+    igraphmodule_handle_igraph_error();
+    Py_DECREF(it);
+    return 1;
+  }
+
+  /* TODO: what about this? is memory managed already? */
+  //IGRAPH_VECTOR_PTR_SET_ITEM_DESTRUCTOR(vec, igraph_vector_int_destroy);
```
szhorvat, line 2717: Yes, fully managed. This can be removed.

src/_igraph/convert.c
iosonofabio: Thanks @szhorvat ! These comments are already quite helpful.

I'm now in the thick of the woods with converting `int -> igraph_error_t` in attribute handling functions, such a nightmare. I'm sure I'll ask a few things about the new types lists, thanks

src/_igraph/convert.c
iosonofabio: yep

src/_igraph/convert.c
iosonofabio: agreed, I called the list `veclist` and the vector `vec`

iosonofabio:
Passes all tests locally, let's see CI. This was quite painless as of now... perhaps too much so. I feel like some major obstacle is looming :smirk:

iosonofabio:
Oh well CI cannot pass until I tell it to use the new version of the C core from `igraph/igraph/develop`... which is a submodule. Let me figure out how to do that...

iosonofabio:
This PR is not urgent but whenever @szhorvat and/or @ntamas have time, you could take a look. Don't expect major issues but perhaps style, memory leaks, or a corner case or two might be off.

src/_igraph/convert.c
```diff
@@ -2584,7 +2584,7 @@ int igraphmodule_PyObject_to_vector_ptr_t(PyObject* list, igraph_vector_ptr_t* v
 
   IGRAPH_VECTOR_PTR_SET_ITEM_DESTRUCTOR(vec, igraph_vector_destroy);
   while ((item = PyIter_Next(it)) != 0) {
-    subvec = igraph_Calloc(1, igraph_vector_t);
+    subvec = IGRAPH_CALLOC(1, igraph_vector_t);
```
szhorvat, line 2587: There is no longer any need to allocate this on the heap. Keep it on the stack.

It seems to me that there was a memory leak in the original function as the heap-allocated vector was not freed on error. https://github.com/igraph/python-igraph/issues/512  In this refactoring, this has been converted to a permanent memory leak (it happens regardless of error).

src/_igraph/convert.c
```diff
@@ -2713,11 +2713,8 @@ int igraphmodule_PyObject_to_vector_list_t(PyObject* list, igraph_vector_list_t*
     return 1;
   }
 
-  /* TODO: what about this? is memory managed already? */
-  //IGRAPH_VECTOR_PTR_SET_ITEM_DESTRUCTOR(vec, igraph_vector_int_destroy);
   while ((item = PyIter_Next(it)) != 0) {
-    /* FIXME: is this even the right way to do this? */
-    subvec = igraph_Calloc(1, igraph_vector_t);
+    subvec = IGRAPH_CALLOC(1, igraph_vector_t);
```
szhorvat: There is no longer any need to allocate this on the heap. Keep it on the stack.

It seems to me that there was a memory leak in the original function as the heap-allocated vector was not freed on error. https://github.com/igraph/python-igraph/issues/512 In this refactoring, this has been converted to a permanent memory leak (it happens regardless of error).

iosonofabio: I think I understand the memory leak, but not the heap/stack talk. What heap, what stack?

szhorvat: instead of

```
igraph_vector_t *subvec;
subvec = IGRAPH_CALLOC(1, igraph_vector_t);
```

use

```
igraph_vector_t subvec;
```

There is no need for heap allocation.

Then there will also be no need for `free`ing what `IGRAPH_CALLOC` allocated.

iosonofabio:
Thanks for looking into memory leaks. I tried to fix the issue here and #512 at the same time, is it how you think it should be or am I freeing too much for the vector pointer (i.e. old data structure) conversions in case no errors happen?

src/_igraph/convert.c
```diff
@@ -2538,11 +2606,12 @@ int igraphmodule_PyObject_to_vector_ptr_t(PyObject* list, igraph_vector_ptr_t* v
     if (igraph_vector_ptr_push_back(vec, subvec)) {
       Py_DECREF(it);
       igraph_vector_destroy(subvec);
+      free(subvec);
       igraph_vector_ptr_destroy_all(vec);
       return 1;
     }
 
-    /* ownership of 'subvec' taken by 'vec' here */
+    /* ownership of 'subvec' taken by 'vec' here, we should still free the pointer */
```
szhorvat, line 2614: Here there is no need to free the pointer.

The code seems correct to me. I'm just commenting your change to the comment.

src/_igraph/convert.c
```diff
@@ -2606,11 +2676,120 @@ int igraphmodule_PyObject_to_vector_int_ptr_t(PyObject* list, igraph_vector_ptr_
     if (igraph_vector_ptr_push_back(vec, subvec)) {
       Py_DECREF(it);
       igraph_vector_int_destroy(subvec);
+      free(subvec);
       igraph_vector_ptr_destroy_all(vec);
       return 1;
     }
 
-    /* ownership of 'subvec' taken by 'vec' here */
+    /* ownership of 'subvec' taken by 'vec' here, we should still free the pointer */
```
szhorvat, line 2684: Same as above: there is no need to free the pointer _here_ as now it belongs to the `vector_ptr_t`. The code is correct, I am referring to the comment only.

szhorvat (commented):
To answer your last question: everything looks good to me @iosonofabio.

I am only looking at the usage of the igraph C library here, as I am not familiar with Python's C API.

ntamas:
FYI, taking a look at this now to let my mind wander off of https://liveuamap.com a bit...

src/_igraph/attributes.c
```diff
@@ -1491,7 +1493,7 @@ static int igraphmodule_i_attribute_get_info(const igraph_t *graph,
     if (n) {
       retval = igraphmodule_PyList_to_strvector_t(keys, n);
       if (retval) {
-        return retval;
+        return (igraph_error_t)retval;
```
ntamas, line 1496: Casting is not correct here; `igraphmodule_PyList_to_strvector_t` returns 0 for success and 1 for error (similarly to many functions in Python's C API). You should do something like `retval ? IGRAPH_FAILURE : IGRAPH_SUCCESS`, which communicates the intention more clearly.

src/_igraph/convert.c
ntamas: Indeed, the code is correct, the comment is not.

src/_igraph/convert.c
ntamas: Agreed.

src/_igraph/convert.c
```diff
@@ -2606,11 +2676,120 @@ int igraphmodule_PyObject_to_vector_int_ptr_t(PyObject* list, igraph_vector_ptr_
...
+  }
+
+  Py_DECREF(it);
+  return 0;
+}
+
+/**
+ * \ingroup python_interface_conversion
+ * \brief Converts a Python list of lists to an \c igraph_vector_list_t
+ *        (containing \c igraph_vector_int_t items).
```
ntamas, line 2694: An `igraph_vector_list_t` contains `igraph_vector_t` items; the comment should be fixed. The code is OK.

src/_igraph/graphobject.c
```diff
@@ -3656,8 +3656,8 @@ PyObject *igraphmodule_Graph_articulation_points(igraphmodule_GraphObject *self)
  */
 PyObject *igraphmodule_Graph_assortativity_nominal(igraphmodule_GraphObject *self, PyObject *args, PyObject *kwds)
 {
-  static char *kwlist[] = { "types", "directed", NULL };
-  PyObject *types_o = Py_None, *directed = Py_True;
+  static char *kwlist[] = { "types", "directed", "normalized", NULL };
```
ntamas, line 3659: If you extend `kwlist` with an additional keyword argument, I think you should also update the corresponding `PyArg_ParseTupleAndKeywords` call; the format string should be `O|OO` and you should add `&normalized` as the last argument, after `&directed`. You have done this below for `igraphmodule_Graph_assortativity()` but not here.

src/_igraph/graphobject.c
```diff
@@ -4633,24 +4633,24 @@ PyObject *igraphmodule_Graph_decompose(igraphmodule_GraphObject * self,
   if (igraphmodule_PyObject_to_connectedness_t(mode_o, &mode))
     return NULL;
 
-  igraph_vector_ptr_init(&components, 3);
+  igraph_graph_list_init(&components, 3);
```
ntamas, line 4636: The return value should be checked here, and you should not initialize the graph list with a length of 3. It was okay for `igraph_vector_ptr_t` because that one would have only created null pointers, but now you are actually creating 3 graphs.

src/_igraph/graphobject.c
```diff
@@ -5017,80 +5016,65 @@ PyObject *igraphmodule_Graph_get_shortest_paths(igraphmodule_GraphObject *
...
     if (weights) { igraph_vector_destroy(weights); free(weights); }
     igraph_vs_destroy(&to);
     igraphmodule_handle_igraph_error();
     return NULL;
   }
 
-  ptrvec = (igraph_vector_ptr_t *) calloc(1, sizeof(igraph_vector_ptr_t));
-  if (!ptrvec) {
+  /* Memory for the pointer */
+  veclist = (igraph_vector_int_list_t *) calloc(1, sizeof(igraph_vector_int_list_t));
```
ntamas, line 5029: I think we don't need to allocate `veclist` on the heap, we can simply use an `igraph_vector_int_list_t`. I have no idea why the original implementation allocated `ptrvec` on the heap; it must have been one of the first functions involving `igraph_vector_ptr_t` that I converted from C to Python and I think I have overcomplicated it.

src/_igraph/graphobject.c
```diff
@@ -5017,80 +5016,65 @@ PyObject *igraphmodule_Graph_get_shortest_paths(igraphmodule_GraphObject *
...
-    free(res);
-    igraph_vector_ptr_destroy(ptrvec); free(ptrvec);
+    igraph_vector_int_list_destroy(veclist); free(veclist);
     if (weights) { igraph_vector_destroy(weights); free(weights); }
     igraph_vs_destroy(&to);
     return NULL;
   }
 
-  igraph_vector_ptr_destroy(ptrvec); free(ptrvec);
-  if (weights) { igraph_vector_destroy(weights); free(weights); }
```
ntamas, line 5069: What happened to `weights`, why are we not destroying it any more?

src/_igraph/graphobject.c
```diff
@@ -5017,80 +5016,65 @@ PyObject *igraphmodule_Graph_get_shortest_paths(igraphmodule_GraphObject *
...
 
+  /* Make empty Python list of paths */
   list = PyList_New(no_of_target_nodes);
   if (!list) {
-    for (j = 0; j < no_of_target_nodes; j++) igraph_vector_int_destroy(&res[j]);
-    free(res);
     return NULL;
   }
 
+  /* Fill list of paths */
```
ntamas, line 5066: We have an `igraph_vector_int_list_to_PyObject()` function in `convert.c`; can't we use that here?

ntamas, line 5066: We have an `igraph_vector_int_list_to_PyObject()` function in `convert.c`; can't we use that here?

src/_igraph/graphobject.c
```diff
@@ -11179,37 +11146,34 @@ PyObject *igraphmodule_Graph_cliques(igraphmodule_GraphObject * self,
...
-  n = igraph_vector_ptr_size(&result);
+  n = igraph_vector_int_list_size(&result);
   list = PyList_New(n);
   if (!list) {
     return NULL;
   }
 
   for (i = 0; i < n; i++) {
-    igraph_vector_int_t *vec = (igraph_vector_int_t *) VECTOR(result)[i];
+    igraph_vector_int_t *vec = igraph_vector_int_list_get_ptr(&result, i);
```
ntamas, line 11166: This bit of code below essentially converts an `igraph_vector_int_list_t` to a Python list-of-tuples. This would be a great addition to `convert.c` and it would simplify the code in the other clique-related functions (e.g. `largest_cliques` or `maximal_cliques`). We don't need to do this as part of this PR, but if you don't, please add an issue for it so we don't forget to do this later (I can do this next week when I'm back on track).

src/_igraph/graphobject.c
```diff
@@ -13373,8 +13322,11 @@ struct PyMethodDef igraphmodule_Graph_methods[] = {
    "  have an out-type and an in-type. In this case, I{types1} contains the\n"
    "  out-types and this parameter contains the in-types in a list or the\n"
    "  name of a vertex attribute. If C{None}, it is assumed to be equal\n"
-   "  to I{types1}.\n\n"
+   "  to I{types1}.\n"
```
ntamas, line 13325: The second `\n` was deliberate; I wanted to add an extra separator line between the function description and the list of parameters.

ntamas (changes requested):
Sorry for the flood of individual comments; it took me a while to realize that I should do a proper review so you get them all at once in a single email.

How are things on the baby front? Let me know if time's up and I'll take over this PR.

src/_igraph/graphobject.c
szhorvat: Just noting that the default must be `normalized=True` on the Python side.

src/_igraph/convert.c
iosonofabio: ok thanks

src/_igraph/attributes.c
iosonofabio: yep, good point

src/_igraph/graphobject.c
iosonofabio: thank you

src/_igraph/graphobject.c
iosonofabio: Please double check the new version here

src/_igraph/graphobject.c
iosonofabio: yep ok

src/_igraph/graphobject.c
iosonofabio: hehe good eye, fixed

src/_igraph/graphobject.c
iosonofabio: yes we can

iosonofabio:
Thanks for the review, I'm going through it.

Can you please clarify when we need to call `igraphmodule_handle_igraph_error()`? It seems quite arbitrary to me, as opposed to raising Python errors (e.g. Memory Exhausted) by hand

src/_igraph/graphobject.c
iosonofabio: ... but isn't that between one parameter and the next?

src/_igraph/graphobject.c
```diff
@@ -5130,7 +5114,7 @@ PyObject *igraphmodule_Graph_get_all_shortest_paths(igraphmodule_GraphObject *
     return NULL;
   }
 
-  if (igraph_vector_ptr_init(&res, 1)) {
+  if (igraph_vector_int_list_init(&res, 1)) {
```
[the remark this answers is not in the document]
iosonofabio: yep

src/_igraph/graphobject.c
```diff
@@ -5151,33 +5135,31 @@ PyObject *igraphmodule_Graph_get_all_shortest_paths(igraphmodule_GraphObject *
   igraph_vs_destroy(&to);
   if (weights) { igraph_vector_destroy(weights); free(weights); }
 
-  IGRAPH_VECTOR_PTR_SET_ITEM_DESTRUCTOR(&res, igraph_vector_int_destroy);
-
-  j = igraph_vector_ptr_size(&res);
+  j = igraph_vector_int_list_size(&res);
```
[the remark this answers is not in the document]
iosonofabio: I think the function has a different name, but yes, let me go through a few cases

src/_igraph/graphobject.c
iosonofabio: Added. Let me examine other places for this situation.

src/_igraph/convert.c
iosonofabio: right, thank you

src/_igraph/graphobject.c
ntamas: Ah yes, you are right, sorry.

ntamas:
Python expects that if you return `NULL` instead of a "real" `PyObject*` from a function, then an exception is set to indicate what the problem is. Normally, we set the exception in `igraphmodule_igraph_error_hook()`, but this is called only through `IGRAPH_CHECK()`. `igraphmodule_handle_igraph_error()` is a safety mechanism to ensure that even if `igraph_...` function did not call `IGRAPH_CHECK()` (but simply returned an igraph error code on its own), then we _still_ have a valid exception that we return to Python. The same applies to `PyErr_SetString(PyExc_MemoryError, "")`.

The easiest would be to call igraph functions _ourselves_ with `IGRAPH_CHECK()`, but we cannot do that because it works only in functions that return `igraph_error_t`. That's why we do `if (igraph_...()) { igraphmodule_handle_igraph_error(); return NULL; }` instead.

src/_igraph/graphobject.c
```diff
@@ -5019,7 +5019,6 @@ PyObject *igraphmodule_Graph_get_shortest_paths(igraphmodule_GraphObject *
   /* Initialize the vector_int_list itself, size is managed internally
    * by the C core function */
   if (igraph_vector_int_list_init(&veclist, 0)) {
-    PyErr_SetString(PyExc_MemoryError, "");
```
ntamas, line 5029: Please add `igraphmodule_handle_igraph_error()` here for sake of consistency.

src/_igraph/graphobject.c
```diff
@@ -11102,6 +11089,7 @@ PyObject *igraphmodule_Graph_cliques(igraphmodule_GraphObject * self,
   }
 
   if (igraph_vector_int_list_init(&res, 0)) {
+    /* FIXME: when is this line needed? */
     PyErr_SetString(PyExc_MemoryError, "");
```
ntamas, line 11093: Yopu can replace this with `igraphmodule_handle_igraph_error()` instead.

ntamas (changes requested):

szhorvat:
> but this is called only through `IGRAPH_CHECK()`.

You mean `IGRAPH_ERROR()`, right?



> even if `igraph_...` function did not call `IGRAPH_CHECK()` (but simply returned an igraph error code on its own),

This should not be allowed. I meant to bring this up at some point, also related to https://github.com/igraph/igraph/pull/1972  We should make it a requirement that a non-success error code can only be returned after having called `igraph_error()` (usually through `IGRAPH_ERROR()`). There are very few cases which don't do this and it will be an easy fix.

ntamas:
> You mean `IGRAPH_ERROR()`, right?

Both. `IGRAPH_CHECK()` also calls `IGRAPH_ERROR()` with an empty reason string.

> > even if igraph_... function did not call IGRAPH_CHECK() (but simply returned an igraph error code on its own),

> This should not be allowed.

Agreed. But this is yet another bureaucratic requirement that would be hard for newcomers to grasp so I'm not sure whether we should really insist on this when processing third-party PRs or just fix it up ourselves after merging.

Nevertheless, we should look for instances of `return IGRAPH_...` in the C core and replace them with `IGRAPH_ERROR()` where possible so we can also provide a meaningful error message along with the error code.

iosonofabio:
Good discussion. I think many new contributors learn by looking at consistent patterns across the codebase and imitating them. So if we use `IGRAPH_CHECK` everywhere in `graphobject.c` and friends, it will feel natural for new PR authors to use it too.

In any case, this has nothing to do with typed lists, and requires a little work on the C side as well. What about I fix the two small things above and then we merge this PR?

After that we go through error codes (@szhorvat 's other proposal in the other repo), more consistent `IGRAPH_ERROR` (as suggested here), both of which are C core issues. When that's done, we can make another PR here to establish good conventions throughout the Python interface.

ntamas:
> So if we use IGRAPH_CHECK everywhere in `graphobject.c` and friends, it will feel natural for new PR authors to use it too.

We can't use `IGRAPH_CHECK` in the Python interface because `IGRAPH_CHECK` expects that it is invoked in a function that returns an `igraph_error_t`. 99% of the functions in `graphobject.c` follow the Python convention where the return value is either a valid `PyObject*` or `NULL`, depending on whether there was an error or not. That's why we don't use `IGRAPH_FINALLY` et al from the Python interface.

We could come up with an `IGRAPH_PY_CHECK` that is similar to `IGRAPH_CHECK` but follows the Python convention. That would clean up things a lot in the Python interface, but it's not the topic of this PR. Let me go through this one once again to see if it is ready for merging now.

iosonofabio:
I was thinking about the `IGRAPH_PY_CHECK`, it would be useful wouldn't it? Unless the next major transition is to an interface generator, of course.

Thanks for merging!

PR #514, opened 2022-02-23
fix: shortest path crash with zero-length weight vector

### LINKED ISSUE #513: Weighted shortest_paths crashes on edgeless graphs.
**Describe the bug**
Running `shortest_paths` with the `weights` parameter on an edgeless graph triggers the fatal error and crashes python: `Fatal error at src/core/vector.c:1181 : Assertion failed: v->stor_begin != v->end`

**To reproduce**
```python
>>> import igraph
>>> g = igraph.Graph()
>>> g.add_vertices(1)
>>> g.es['attr'] = []
>>> g.shortest_paths(weights='attr')
```

**Version information**
Crashes in 0.9.9.
Also tested in 0.7.1.post6, everything works fine.

szhorvat:
Fixes #513.

This is trivial, but I'm not yet comfortable directly committing to python-igraph

### DIFF
--- a/src/_igraph/graphobject.c
+++ b/src/_igraph/graphobject.c
@@ -5826,7 +5826,7 @@ PyObject *igraphmodule_Graph_shortest_paths(igraphmodule_GraphObject * self,
       }
     }
   } else {
-    /* No weights, use a simple BFS */
+    /* No weights or empty weight vector, use a simple BFS */
     e = igraph_shortest_paths(&self->g, &res, from_vs, to_vs, mode);
   }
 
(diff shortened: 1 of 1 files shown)


szhorvat:
I checked, and could not find any other similar problems in the python-igraph codebase.

iosonofabio:
Would you mind changing the comment of line 5833 into:

```
/* No weights or empty weight vector, use a simple BFS */
```

iosonofabio:
Then you can merge

szhorvat:
Merged. I'll leave the `master` -> `develop` merge to you, in case some conflict appears that I wouldn't be comfortable resolving.

ntamas:
Merged into `develop`

PR #529, opened 2022-04-15
Add umap function

iosonofabio:
Exporting UMAP to the Python interface.

TODO:
- [x] Update submodule hash
- [x] Add tests
- [x] Add docs
- [x] Polish:
  - [x] Better defaults: the C layer has no defaults

### DIFF
--- a/src/_igraph/graphobject.c
+++ b/src/_igraph/graphobject.c
@@ -7814,12 +7814,13 @@ PyObject *igraphmodule_Graph_layout_umap(igraphmodule_GraphObject * self,
       return NULL;
     }
   } else {
-    /* FIXME: complain about the dimension */
     if (dist) {
       igraph_vector_destroy(dist); free(dist);
     }
     igraph_matrix_destroy(&m);
+    PyErr_SetString(PyExc_ValueError, "dim must be 2 or 3");
     igraphmodule_handle_igraph_error();
+    return NULL;
   }
 
   if (dist) {
@@ -15058,9 +15059,9 @@ struct PyMethodDef igraphmodule_Graph_methods[] = {
    "@param epochs: the number of epochs (iterations) the algorithm will iterate over.\n"
    "  Notice that UMAP does not technically converge for symmetry reasons, but a \n"
    "  larger number of epochs should generally give an equivalent or better layout.\n"
-   "@sampling_prob: the probability of sampling each vertex for repulsion at each\n"
-   "  epoch or iteration. A higher probability will give better results but also\n"
-   "  require more computations.\n"
+   "@param sampling_prob: the probability of sampling each vertex for repulsion at\n"
+   "  each epoch or iteration. A higher probability will give better results but\n"
+   "  also require more computations. Values between 50 and 1000 are typical.\n"
    "@return: the calculated layout.\n\n"
    "@newfield ref: Reference\n"
    "@ref: L McInnes, J Healy, J Melville: UMAP: Uniform Manifold Approximation \n"
--- a/src/_igraph/graphobject.c
+++ b/src/_igraph/graphobject.c
@@ -7751,7 +7751,7 @@ PyObject *igraphmodule_Graph_layout_umap(igraphmodule_GraphObject * self,
                                         PyObject * args, PyObject * kwds)
 {
   static char *kwlist[] =
-    { "dist", "dim", "min_dist", "epochs", "sampling_prob", NULL };
+    { "dist", "dim", "seed", "min_dist", "epochs", "sampling_prob", NULL };
   igraph_matrix_t m;
   igraph_vector_t *dist = 0;
   Py_ssize_t dim = 2;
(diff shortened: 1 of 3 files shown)


src/_igraph/graphobject.c
```diff
@@ -7740,6 +7740,104 @@ PyObject *igraphmodule_Graph_layout_sugiyama(
...
+
+  if (sampling_prob <= 0) {
+    /* FIXME: How to actually tell the user? */
+    igraphmodule_handle_igraph_error();
+  }
+
+  if (sampling_prob > 1) {
+    /* FIXME: How to actually tell the user? */
+    igraphmodule_handle_igraph_error();
+  }
```
szhorvat, line 7786: These kinds of checks are normally done only in the C core and not duplicated in the high-level interface.

src/_igraph/graphobject.c
```diff
@@ -7740,6 +7740,104 @@ PyObject *igraphmodule_Graph_layout_sugiyama(
...
+
+  if (dist_o != Py_None) {
+    dist = (igraph_matrix_t*)malloc(sizeof(igraph_matrix_t));
```
szhorvat, line 7789: What is the purpose of allocating this dynamically? @ntamas I can see this pattern throughout this file, but isn't it just complicating the code and providing one more chance to forget `free`? We just fixed a bunch of leaks like this.

src/_igraph/graphobject.c
iosonofabio: Thanks. What about `CHECK_SSIZE_T_RANGE_POSITIVE` the lines right above?

src/_igraph/graphobject.c
iosonofabio: Good question... @ntamas?

src/_igraph/graphobject.c
iosonofabio: Okay I removed all of that, it's checked later on anyway.

szhorvat: `CHECK_SSIZE_T_RANGE_POSITIVE` is Python-specific so it should stay. It checks that the integer coming from Python fits into an `igraph_integer_t`.

To be clear, I'm not at all experienced with the Python interface, and part of the reason I'm looking at this is to better understand how to expose functions to Python.

src/_igraph/graphobject.c
ntamas: Yes, that's unnecessary, you could allocate it on the stack. I was inexperienced when I initially wrote code like this, and then it got copied all over the place.

iosonofabio: no problem. Would you like me to fix these here throughout?

iosonofabio:
Okay, it compiles and the basic tests (copied from the C layer) pass.

I'll have to improve some function defaults and their documentation, improve docs in general, fix things like the stack/heap memory issue above and polish the whole thing.

Should be ready for review by Easter I reckon

tests/test_layouts.py
```diff
@@ -264,21 +266,26 @@ def testUMAP(self):
...
+                InternalError, g.layout_umap,
                 sampling_prob=-0.01,
             )
 
-        self.assertRaises(
-                ValueError, g.layout_umap,
-                sampling_prob=1.01,
-            )
+        # TODO: this is currently accepted by the C core function...
+        # document at the Python level or fix at the C level?
```
szhorvat, line 274: Invalid parameter values should be rejected at the C core level. I don't want to have to re-discover these problems and re-fix them in the Mathematica interface.

iosonofabio: agreed, let me fix it in C then

src/_igraph/graphobject.c
ntamas: one thing at a time :) we can do it outside the scope of this PR. Feel free to add an issue for it and then I'll do it eventually - or someone else can, but let's not mix more things in here

ntamas: FWIW I checked the code more carefully and it seems like this pattern is used for optional input arguments; for instance, in the interface for `layout_mds()`, the user _may_ provide a distance matrix but it is not necessary. Currently the pattern is that `dist == 0` if the user did not provide a matrix, and `dist != 0` means that the user did provide one in Python, in which case we allocate a corresponding C-level matrix on the heap. We could pre-allocate one on the stack and then we can get rid of the `free()` call, but then we need a `p_dist` pointer that is either `NULL` (if the user did not provide a distance matrix in Python) or points to the pre-allocated matrix if needed.

There are maybe 4 or 5 similar occasions in `graphobject.c` so I don't think it's a big deal. If one of you decides to simplify it by using a pre-allocated matrix, please do it in a separate PR so it is reviewed more easily.

src/_igraph/graphobject.c
```diff
@@ -7742,6 +7742,95 @@ PyObject *igraphmodule_Graph_layout_sugiyama(
...
+      igraphmodule_handle_igraph_error();
+      return NULL;
+    }
+  } else {
+    /* FIXME: complain about the dimension */
+    if (dist) {
+      igraph_vector_destroy(dist); free(dist);
+    }
+    igraph_matrix_destroy(&m);
+    igraphmodule_handle_igraph_error();
```
ntamas, line 7822: This won't work here because igraph did not raise any error if we managed to get to this point. You need to do a `PyErr_SetString(PyExc_ValueError, "dim must be 2 or 3")` or something like that.

src/_igraph/graphobject.c
```diff
@@ -14952,6 +15041,31 @@ struct PyMethodDef igraphmodule_Graph_methods[] = {
...
+   "@param dist: distances associated with the graph edges. If None, all edges will\n"
+   "  be assumed to convey the same distance between the vertices.\n"
+   "@param dim: the desired number of dimensions for the layout. dim=2\n"
+   "  means a 2D layout, dim=3 means a 3D layout.\n"
+   "@param min_dist: the minimal distance in the embedded space beyond which the\n"
+   "  probability of being located closeby decreases.\n"
+   "@param epochs: the number of epochs (iterations) the algorithm will iterate over.\n"
+   "  Notice that UMAP does not technically converge for symmetry reasons, but a \n"
+   "  larger number of epochs should generally give an equivalent or better layout.\n"
+   "@sampling_prob: the probability of sampling each vertex for repulsion at each\n"
```
ntamas, line 15061: Should be `@param sampling_prob`

tests/test_layouts.py
```diff
@@ -250,6 +251,74 @@ def testMDS(self):
         lo = g.layout("mds")
         self.assertTrue(isinstance(lo, Layout))
 
+    def testUMAP(self):
+        g = Graph()
+
+        # These distinct types of exceptions look ugly... can we improve?
```
ntamas, line 257: At some point some of the catch-all igraph `InternalError` exceptions should be converted to proper Python exceptions, but that's API-breaking and I'm not sure whether there's a straightforward mapping that makes sense in most contexts. The only information we can use in our error handler to decide whether we want to convert to a Python exception is the igraph error code, and that may not always be enough.

Let's leave it to the subject of another issue / PR.

tests/test_layouts.py
```diff
@@ -250,6 +251,74 @@ def testMDS(self):
...
+            0.1, 0.09, 0.12, 0.09, 0.1, 0.1,
+            0.9, 0.9, 0.9,
+            0.2, 0.1, 0.1, 0.1, 0.1, 0.1, 0.08, 0.05, 0.1, 0.08, 0.12, 0.09, 0.11
+            ]
+        g = Graph(edges)
+        lo = g.layout_umap(dist=dist, epochs=500, sampling_prob=0.3)
+        self.assertTrue(isinstance(lo, Layout))
+
+        # One should get two clusters in this case
+        xmax = max([x for (x, y) in lo.coords])
```
ntamas, line 306: `max` and `min` work with generator expressions so you can omit the square brackets here and in the next three lines.

tests/test_layouts.py
```diff
@@ -250,6 +251,74 @@ def testMDS(self):
...
+        xmin = min([x for (x, y) in lo.coords])
+        ymin = min([y for (x, y) in lo.coords])
+        distmax = max(xmax - xmin, ymax - ymin)
+        for iclu in range(0, 8, 7):
+            xclu = sum([x for (x, y) in lo.coords[iclu:iclu + 4]]) / 4
+            yclu = sum([y for (x, y) in lo.coords[iclu:iclu + 4]]) / 4
+            for i in range(4):
+                dx = lo.coords[iclu + i][0] - xclu
+                dy = lo.coords[iclu + i][1] - yclu
+                dist = sqrt(dx * dx + dy * dy)
```
ntamas, line 317: Use `math.hypot()` because that is less susceptible to overflows (and it also works in higher dimensions).

ntamas (changes requested):
Some minor suggestions here and there.

iosonofabio:
I'm opening another small PR at the C level to add an argument `bool skip_initialization`, which allows users to start the UMAP minimization with a custom layout (instead of random or possibly spectral). I think that's especially useful for layout perturbations (e.g. a new node is added) that would be run with few epochs.

Assuming that's merged soon, I'll propagate the argument up here and the PR should be then ready for merging.

ntamas:
LGTM so far. I'll take a look at the C side as well.

szhorvat:
This parameter is named `use_seed` in other layout functions. Can we make it consistent with them?

iosonofabio:
uh-oh, the C level API for random walks changed and that screws this PR. Let me open a separate PR about that and this will have to queue right after...

iosonofabio:
Alright, this depends on #532 which itself depends on #533 unless we find the bug by hand quicker

iosonofabio:
I still get test failures on the random walks on my machine... the UMAP tests pass though, which is nice :smirk: 

Let me see if I can understand what's going on.... perhaps I merged #532 too quickly

iosonofabio:
Actually, just me being now very smart... it should work now that I updated the submodule

ntamas:
Here it goes, let me know whether it works in practice :)

PR #532, opened 2022-04-21
Fix random_walk change in C-level API

iosonofabio:
`igraph_random_walk`'s API has changes in C, so this is an effort to adapt the Python interface.

not sure what the current C-level segfaults are caused by...

### DIFF
--- a/src/_igraph/convert.c
+++ b/src/_igraph/convert.c
@@ -3825,4 +3825,3 @@ int igraphmodule_PyObject_to_random_tree_t(PyObject *o, igraph_random_tree_t *re
 
   return igraphmodule_PyObject_to_enum_strict(o, random_tree_tt, (int*)result);
 }
-
--- a/src/_igraph/graphobject.c
+++ b/src/_igraph/graphobject.c
@@ -12155,16 +12155,26 @@ PyObject *igraphmodule_Graph_community_leiden(igraphmodule_GraphObject *self,
  */
 PyObject *igraphmodule_Graph_random_walk(igraphmodule_GraphObject * self,
   PyObject * args, PyObject * kwds) {
-  static char *kwlist[] = { "start", "steps", "mode", "stuck", NULL };
-  PyObject *start_o, *mode_o = Py_None, *stuck_o = Py_None, *res;
+  static char *kwlist[] = { "start", "steps", "mode", "stuck", "weights", "return_type", NULL };
+  PyObject *start_o, *mode_o = Py_None, *stuck_o = Py_None, *resv, *rese, *res;
   igraph_integer_t start;
   Py_ssize_t steps = 10;
   igraph_neimode_t mode = IGRAPH_OUT;
   igraph_random_walk_stuck_t stuck = IGRAPH_RANDOM_WALK_STUCK_RETURN;
-  igraph_vector_int_t walk;
+  igraph_vector_t *ws=0;
+  PyObject *weights_o = Py_None;
+  igraph_vector_int_t vertices, edges;
+  static igraphmodule_enum_translation_table_entry_t return_type_tt[] = {
+        {"vertices", 1},
+        {"edges", 2},
+        {"both", 3},
+        {0,0}
+    };
+  int return_type=1;
+  PyObject *return_type_o = Py_None;
 
-  if (!PyArg_ParseTupleAndKeywords(args, kwds, "|OnOO", kwlist, &start_o,
-        &steps, &mode_o, &stuck_o))
+  if (!PyArg_ParseTupleAndKeywords(args, kwds, "|OnOOOO", kwlist, &start_o,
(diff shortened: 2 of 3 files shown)


szhorvat:
I haven't looked at the code, just a tip: build with address sanitizer to see why it crashes. I wrote up detailed instructions a while ago. https://github.com/igraph/igraph/wiki/Using-sanitizers-to-find-bugs

szhorvat:
It would actually be very good to set up a CI workflow that runs the py-igraph test suite with ASan and LSan. However, we need to use a virtualenv which does not have all the optional Python dependencies (especially matplotlib) because LSan finds many leaks in them. We also need to make sure that the test suite passes without optional dependencies (i.e. these tests are just skipped). I think Tamás fixed this recently.

@iosonofabio If you can help me with the Python side, we can set this up together, in a separate PR, then return to this one.

iosonofabio:
I tried following instructions for the sanirizers, and I got:

```
[python-igraph] (random_walk_fix)$ .venv/bin/pytest
============================================================== test session starts ==============================================================
platform linux -- Python 3.10.4, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /home/fabio/university/PI/projects/python-igraph
collected 462 items                                                                                                                             

tests/test_atlas.py .........                                                                                                             [  1%]
tests/test_attributes.py ..............................                                                                                   [  8%]
tests/test_basic.py ....................................................                                                                  [ 19%]
tests/test_bipartite.py ........                                                                                                          [ 21%]
tests/test_cliques.py ............                                                                                                        [ 24%]
tests/test_colortests.py ...........                                                                                                      [ 26%]
tests/test_conversion.py .........                                                                                                        [ 28%]
tests/test_decomposition.py ....................................                                                                          [ 36%]
tests/test_edgeseq.py ................................                                                                                    [ 43%]
tests/test_flow.py .................                                                                                                      [ 46%]
tests/test_foreign.py ......s.....s........                                                                                               [ 51%]
tests/test_games.py ...........                                                                                                           [ 53%]
tests/test_generators.py .................                                                                                                [ 57%]
tests/test_homepage.py ....                                                                                                               [ 58%]
tests/test_indexing.py ....                                                                                                               [ 59%]
tests/test_isomorphism.py ...........                                                                                                     [ 61%]
tests/test_iterators.py .....                                                                                                             [ 62%]
tests/test_layouts.py ......................                                                                                              [ 67%]
tests/test_matching.py .....                                                                                                              [ 68%]
tests/test_operators.py ...............................                                                                                   [ 75%]
tests/test_rng.py ...                                                                                                                     [ 75%]
tests/test_separators.py .....                                                                                                            [ 76%]
tests/test_spectral.py ..                                                                                                                 [ 77%]
tests/test_structural.py ................................................                                                                 [ 87%]
tests/test_unicode_issues.py ..                                                                                                           [ 88%]
tests/test_vertexseq.py .............................                                                                                     [ 94%]
tests/test_walks.py Fatal Python error: Segmentation fault

Thread 0xSegmentation fault (core dumped)
```

which does not seem terribly useful.

Re separate PR, sure. I mean, this PR is already a branch off the UMAP PR which is what I really care about, but having sanitizers in CI would be useful so let's do it.

szhorvat:
That is strange, as if sanitizers were not active at all...

Can you set up a CI workflow (in a PR) that will:

 - Build py-igraph
 - Install it into a venv with only the required dependencies, and nothing else (no matplotlib/cairo/numpy/networkx/whatever)
 - Runs the test suite

It should not do anything else. I'll try to add the sanitizer stuff to it afterwards.

I'm not entire comfortable with Python development workflows, so it would be really helpful to have this starting point.

iosonofabio:
no problem, let me set it up. I might have messed up on my end anyway

iosonofabio:
Yep, created in #533

ntamas:
I usually simply use `pytest -s` to disable stdout/stderr capturing completely.

szhorvat:
Haven't looked at your PR in detail, but note that the interpretation of `steps` has changed for the edge walk (out of necessity when coupling with the vertex walk).

ntamas:
I'm investigating; this might actually be a bug in the C core. The returned vertex vector contains a duplicate vertex; this is what I get when seeding the test case to make it deterministic:

```
{'edges': [23, 21, 7], 'vertices': [11, 3, 3, 1]}
```

ntamas:
> Haven't looked at your PR in detail, but note that the interpretation of steps has changed for the edge walk (out of necessity when coupling with the vertex walk).

I think that's okay; the parameter is called `steps` on the Python side and the docs say `the number of steps that the random walk should take`, which is correct no matter whether you want a vertex or an edge walk.

ntamas:
This looks good to me now from the partial PR results. @iosonofabio if you are satisfied with my changes and all tests pass, feel free to squash-and-merge.

PR #533, opened 2022-04-21
Address sanitizer ci

iosonofabio:
Make a CI build that uses sanitizers to help debugging at the C-Python interface

@szhorvat: happy to work on this together, just push to the branch whenever you want.

Instructions from here:

https://github.com/igraph/igraph/wiki/Using-sanitizers-to-find-bugs

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -280,4 +280,4 @@ jobs:
           LD_PRELOAD: "/lib/x86_64-linux-gnu/libasan.so.5:/lib/x86_64-linux-gnu/libubsan.so.1"
           ASAN_OPTIONS: "detect_leaks=0"
         run: |
-          python -m pytest tests
+          python -m pytest --capture=sys tests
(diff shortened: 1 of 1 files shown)


szhorvat:
OK, I'm done with this for now.

 - Leak detection is disabled because apparently Python leaks (as far as I can tell from their issue tracker, this might be by design). I do not understand why I see this only on Linux and not macOS.
 - We compile with NDEBUG to disable `assert`s in RNG functions. There is a bug in the C core, triggered by the Python test suite, where uniform random numbers are requested in a way that the upper bound is smaller than the lower bound. It's in one of the layout functions, as I recall FR. I added this assert myself recently because the function does not work correctly if `h < l`. I'm not sure if it's even a good idea to allow this because normally it is expected that numbers are generated from `[l, h)`, i.e. excluding `h`.

szhorvat:
I'm not sure if using pytest is a good idea. It suppresses output, so we can't see the messages from ASan. I don't know how to fix this, but we could simply use `python -m unittest` instead.

vtraag (commented):
The workflow overall looks good to me, I don't have any specific remarks.

I should say though that I have little experience building with sanitizers, so I can't say anything about that specifically.

szhorvat:
@iosonofabio I suggest you merge this and make use of it for debugging #532.

Issue #566, opened 2022-09-08
Cannot convert networkx graph to igraph using `ig.Graph.from_networkx`


mcleantom:
**Describe the bug**
Code which used `ig.Graph.from_networkx` worked in `0.9.11` fails in `0.10.0`

**To reproduce**
Steps or minimal example code to reproduce the problem.

Run the following code:

```python
import networkx as nx
import igraph as ig
import numpy as np

coords = np.array(
    [
        [-7.25, 49.42],
        [-7.99451822, 49.40710344],
        [-8.73857413, 49.38941589],
        [-9.48202378, 49.36694374],
        [-10.22472414, 49.33969506],
        [-10.96653322, 49.30767965],
        [-11.70731035, 49.27090898],
        [-12.44691629, 49.22939616],
        [-13.18521351, 49.18315596],
        [-13.92206632, 49.13220475],
        [-14.65734104, 49.07656048],
    ]
)
speeds = np.ones((len(coords),))
bearings = np.ones_like(speeds)

G = nx.DiGraph()
G.add_edges_from(
    (
        (
            (0, 0),
            (1, 1),
            {
                "weight": float("inf"),
                "coords": coords,
                "speed": speeds,
                "bearing": bearings,
                "delta_time": 1,
            },
        ),
        (
            (0, 0),
            (1, 0),
            {
                "weight": float("inf"),
                "coords": coords,
                "speed": speeds,
                "bearing": bearings,
                "delta_time": 1,
            },
        ),
        (
            (0, 0),
            (1, -1),
            {"weight": 1, "coords": coords, "speed": speeds, "bearing": bearings, "delta_time": 1},
        ),
        (
            (1, 1),
            (2, 0),
            {"weight": 1, "coords": coords, "speed": speeds, "bearing": bearings, "delta_time": 1},
        ),
        (
            (1, 0),
            (2, 0),
            {"weight": 1, "coords": coords, "speed": speeds, "bearing": bearings, "delta_time": 1},
        ),
        (
            (1, -1),
            (2, 0),
            {"weight": 1, "coords": coords, "speed": speeds, "bearing": bearings, "delta_time": 1},
        ),
        (
            (2, 0),
            (3, 0),
            {"weight": 3, "coords": coords, "speed": speeds, "bearing": bearings, "delta_time": 1},
        ),
    )
)
igG = ig.Graph.from_networkx(G)
```

And the following error occurs:

```
[...]

PR #567, opened 2022-09-08
networkx g.nodes is not a list but is iterable

iosonofabio:
Aiming at fixing #566.

We should add a test to this one before merging.

### DIFF
--- a/src/igraph/io/libraries.py
+++ b/src/igraph/io/libraries.py
@@ -98,7 +98,7 @@ def _construct_graph_from_networkx(cls, g, vertex_attr_hashable : str = "_nx_nam
     vcount = len(vnames)
 
     # Dictionary connecting networkx hashables with igraph indices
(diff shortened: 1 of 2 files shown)


ntamas:
Okay, so, if I understand correctly, the purpose of that condition would be to check whether all nodes have an attribute named `_igraph_index` in the NetworkX graph, right? I don't think that `next(iter(g.nodes))` (as suggested in this PR) would do the trick as `g.nodes` seems to iterate over the node _names_.

I think the correct fix would be `next(iter(nxg.nodes.values()))`. This is a bit convoluted, but basically `nxg.nodes` returns a dict-like view that maps NetworkX node names to their "data dictionaries"; `next(iter(...))` then takes the first element of it, which is the data dictionary of the first node.

iosonofabio:
yep... shall we squash & merge?

PR #577, opened 2022-09-29
Integrating igraph_get_k_shortest_paths with python-igraph

sombreslames:
Hello,
This merge request aims to render available within python-igraph the C function igraph_get_k_shortest_paths with python-igraph.

### DIFF
--- a/src/_igraph/graphobject.c
+++ b/src/_igraph/graphobject.c
@@ -5139,13 +5139,13 @@ PyObject *igraphmodule_Graph_get_k_shortest_paths(igraphmodule_GraphObject *
                                                     self, PyObject * args,
                                                     PyObject * kwds)
 {
-  static char *kwlist[] = {"k", "v", "to", "weights", "mode", NULL };
+  static char *kwlist[] = {"k", "from", "to", "weights", "mode", NULL };
   igraph_vector_int_list_t res;
   igraph_vector_t *weights = 0;
   igraph_neimode_t mode = IGRAPH_OUT;
   igraph_integer_t from;
   igraph_integer_t k;
-  igraph_vs_t to;
+  igraph_integer_t to;
   PyObject *list, *from_o, *mode_o=Py_None, *to_o=Py_None, *weights_o=Py_None,*k_o=Py_None;
 
   if (!PyArg_ParseTupleAndKeywords(args, kwds, "O|OOO", kwlist, &k_o, &from_o,
@@ -5158,10 +5158,10 @@ PyObject *igraphmodule_Graph_get_k_shortest_paths(igraphmodule_GraphObject *
   if (igraphmodule_PyObject_to_integer_t(k_o, &k))
     return NULL;
 
-  if (igraphmodule_PyObject_to_vid(from_o, &from, &self->g))
+  if (igraphmodule_PyObject_to_integer_t(from_o, &from))
     return NULL;
 
-  if (igraphmodule_PyObject_to_vs_t(to_o, &to, &self->g, 0, 0))
+  if (igraphmodule_PyObject_to_integer_t(to_o, &to))
     return NULL;
 
   if (igraphmodule_attrib_to_vector_t(weights_o, self, &weights,
@@ -14202,11 +14202,8 @@ struct PyMethodDef igraphmodule_Graph_methods[] = {
    "get_all_shortest_paths(k, v, to=None, weights=None, mode=\"out\")\n--\n\n"
    "Calculates all of the shortest paths from/to a given node in a graph.\n\n"
    "@param k: the number of shortest path desired\n"
-   "@param v: the source for the calculated paths\n"
-   "@param to: a vertex selector describing the destination for\n"
-   "  the calculated paths. This can be a single vertex ID, a list of\n"
-   "  vertex IDs, a single vertex name, a list of vertex names or a\n"
-   "  L{VertexSeq} object. C{None} means all the vertices.\n"
+   "@param from: the ID of the vertex from which the paths are calculated.\n"
+   "@param to: the ID of the vertex to which the paths are calculated.\n"
    "@param weights: edge weights in a list or the name of an edge attribute\n"
    "  holding edge weights. If C{None}, all edges are assumed to have\n"
    "  equal weight.\n"
--- a/src/_igraph/graphobject.c
+++ b/src/_igraph/graphobject.c
@@ -5130,6 +5130,68 @@ PyObject *igraphmodule_Graph_get_all_shortest_paths(igraphmodule_GraphObject *
   igraph_vector_int_list_destroy(&res);
   return list ? list : NULL;
 }
+/** \ingroup python_interface_graph
+ * \brief Calculates the k-shortest paths from/to a given node in the graph
+ * \return a list containing the k-shortest paths from/to the given node
+ * \sa TODO I don't know what to write here : igraph_get_shortest_paths
+ */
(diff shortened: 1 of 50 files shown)


iosonofabio:
There you go!

Let's see if it passes CI. Il give you a few suggestions after that, but it generally looks pretty good 😊

sombreslames:
I did some last update for correct typing should be a bit better.
Where do I write the python doc however ?

iosonofabio:
We're getting there. Two suggestions:

1. CI currently segfaults, take a look and let me know if you want me to check your code
2. Line 14202 and following are the Python docstring. We can polish it once CI passes and the code itself is polished. The lines above the function declaration are internal documentation, that is useful for future us when we will forget what that function does 😋

szhorvat:
We should set up AddressSanitizer runs for py-igraph so it's easier to track down these crashes.

sombreslames:
Shall I split the test to better identify the error ? 
I could also try to install the testing framework. However I have windows, I could try it on docker.

src/_igraph/graphobject.c
```diff
@@ -5130,6 +5130,71 @@ PyObject *igraphmodule_Graph_get_all_shortest_paths(igraphmodule_GraphObject *
   igraph_vector_int_list_destroy(&res);
   return list ? list : NULL;
 }
+/** \ingroup python_interface_graph
+ * \brief Calculates the k-shortest paths from/to a given node in the graph
+ * \return a list containing the k-shortest paths from/to the given node
+ * \sa TODO I don't know what to write here : igraph_get_shortest_paths
```
ntamas, line 5136: Just leave it out, we don't need these references any more :)

src/_igraph/graphobject.c
```diff
@@ -5130,6 +5130,71 @@ PyObject *igraphmodule_Graph_get_all_shortest_paths(igraphmodule_GraphObject *
...
+        &to_o, &weights_o, &mode_o))
+    return NULL;
+
+  if (igraphmodule_PyObject_to_neimode_t(mode_o, &mode))
+    return NULL;
+
+  if (igraphmodule_PyObject_to_integer_t(k_o, &k))
+    return NULL;
+
+  if (igraphmodule_PyObject_to_integer_t(from_o, &from))
```
ntamas, line 5161: This should be `igraphmodule_PyObject_to_vid` because `from` is a vertex index, but this conversion function also allows the user to use a vertex _name_ and it will properly be resolved to the corresponding index.

src/_igraph/graphobject.c
```diff
@@ -5130,6 +5130,71 @@ PyObject *igraphmodule_Graph_get_all_shortest_paths(igraphmodule_GraphObject *
...
+    return NULL;
+
+  if (igraphmodule_PyObject_to_integer_t(to_o, &to))
```
ntamas, line 5164: This should be `igraphmodule_PyObject_to_vid` because `to` is a vertex index, but this conversion function also allows the user to use a vertex _name_ and it will properly be resolved to the corresponding index.

src/_igraph/graphobject.c
```diff
@@ -5130,6 +5130,71 @@ PyObject *igraphmodule_Graph_get_all_shortest_paths(igraphmodule_GraphObject *
...
+    return NULL;
+
+  if (igraphmodule_attrib_to_vector_t(weights_o, self, &weights,
+      ATTRIBUTE_TYPE_EDGE)) {
+    igraph_vs_destroy(&to);
```
ntamas, line 5169: `to` is an `igraph_integer_t` so you should not call `igraph_vs_destroy` on it.
(Or, if the function requires an `igraph_vs_t` as the input, then you need to use `igraphmodule_PyObject_to_vs_t` to convert the input into an `igraph_vs_t`.

ntamas (changes requested):

ntamas:
I think the reason for the CI failures is the call to `igraph_vs_destroy(&to)`, where `to` is an `igraph_integer_t` and not an `igraph_vs_t`. I guess this was accidentally copied from a function that supported using multiple target vertices (and hence used an `igraph_vs_t` as the type of `to` and not an `igraph_integer_t`).

ntamas:
Sorry, I did not mean to mark it as ready for review without your approval, I just accidentally clicked on it.
CI is running again.

ntamas:
Now this seems interesting:

```
>>> g = Graph.Lattice([5, 5], circular=False)
>>> sps = sorted(g.get_k_shortest_paths(6, 0, 12))
igraph._igraph.InternalError: Error at src/misc/other.c:368: Edge IDs do not form a continuous path. -- Invalid value
```

`igraph_get_k_shortest_paths()` priamrily works with edge IDs; if you ask for a path consisting of vertex IDs, it attempts to convert the sequence of edge IDs in the obtained path to vertex IDs. The error indicates that there could be a bug in the C core; we'll need to investigate this separately.

sombreslames:
Ok, no more segfault, just a test that does not work. I will investigate but it seems curious.

sombreslames:
> Now this seems interesting:
> 
> ```
> >>> g = Graph.Lattice([5, 5], circular=False)
> >>> sps = sorted(g.get_k_shortest_paths(6, 0, 12))
> igraph._igraph.InternalError: Error at src/misc/other.c:368: Edge IDs do not form a continuous path. -- Invalid value
> ```
> 
> `igraph_get_k_shortest_paths()` priamrily works with edge IDs; if you ask for a path consisting of vertex IDs, it attempts to convert the sequence of edge IDs in the obtained path to vertex IDs. The error indicates that there could be a bug in the C core; we'll need to investigate this separately.

Ok, do you need anything from me now?
I am afraid that investing the issue in the C core is a bit out of my abilities.
Thanks for the help anyway and have a good day.

ntamas:
I will take a look at this issue to see if it's reproducible in the C core (without going through Python), and then report back here.

szhorvat:
It may be a good test in the C core to compare this function with `igraph_get_all_simple_paths()`, and make sure they give the same result. This is only possible in the unweighted case.

szhorvat:
> We should set up AddressSanitizer runs for py-igraph so it's easier to track down these crashes.

This is in fact set up already. In fact I contributed to setting it up, but I forgot ...

ntamas:
The bug that triggered the `igraph._igraph.InternalError` in the tests should now be fixed in the latest version of the C core. I have brought the PR up-to-date with `master`; let's see if the tests pass now.

ntamas:
Getting close now! :) A few minor comments from my end:

* I'd rather prefer the order of arguments to be `v`, `to`, `k`, `weights` and `mode`. This is because a reasonable default value for `k` is 1, and you cannot put arguments with default values in front of arguments without default arguments, so `k` has to come after `to`.
* I'd like to have an `output` parameter for `get_k_shortest_paths`, similarly to how it works for `get_shortest_paths`, for sake of consistency. `output` must be `"vpath"` if the user wants vertex IDs in the result and `"epath"` if the user wants edge IDs. The default is `"vpath"`. I have refactored the handling of this argument from `igraphmodule_Graph_get_shortest_paths()` into a separate function so you only need to call the same function from `igraphmodule_Graph_get_k_shortest_paths()` to achieve the same thing.

Let me know if you don't have time to fix these in the next few days; I'm aiming for a new release early next week and it would be nice to get this into the next release, so if you are busy with other things, I will take care of it on my own.

sombreslames:
I changed the order ofthe arg, set K by default to 1 and added the output arg

ntamas:
Made a few changes here and there; I'll merge this tomorrow if the CI tests pass.

ntamas:
Thanks a lot for your contribution!

PR #578, opened 2022-09-30
doc: Corrected name of argument.

vtraag:
This corrects the name of the argument for `igraph.Graph.Tree`. The deprecated name was `type`, the new name is `mode`, but this was not yet changed in the docs.


ntamas:
Thanks!

PR #579, opened 2022-09-30
Update cibuildwheel, add dependabot.

vtraag:
This is an attempt to update the `cibuildwheel` to the "official" PyPA version. It also adds a ["dependabot"](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions), which automatically updates all dependencies in the GH Actions.


src/_igraph/graphobject.c
```diff
@@ -13411,7 +13411,7 @@ struct PyMethodDef igraphmodule_Graph_methods[] = {
    "Generates a tree in which almost all vertices have the same number of children.\n\n"
    "@param n: the number of vertices in the graph\n"
    "@param children: the number of children of a vertex in the graph\n"
-   "@param type: determines whether the tree should be directed, and if\n"
+   "@param mode: determines whether the tree should be directed, and if\n"
```
ntamas, line 13414: Does this change also belong to this PR?

vtraag:
Ah, no, sorry that was part of my previous PR that you already merged.

src/_igraph/graphobject.c
vtraag: Ah, no, sorry that was part of my previous PR that you already merged.

ntamas:
Looks like we have a genuine problem with reference counting on Python 3.11, which will be hard to debug but we need to bite the bullet at some point. Failures related to refcounting on Python 3.11 should not prevent this PR from being merged.

vtraag:
The ASan build doesn't reveal much more information. What would you like to do? Would you like to merge this while disabling the builds for 3.11 for the moment and opening a separate issue for the refcount problem? Or would your prefer to leave this PR open until we have solved the refcount problem?

PR #590, opened 2022-10-07
docs: universal way to refer to igraph config file location on Windows

szhorvat:
Because the location of the user directory is not always the same on all Windows systems. The one I am dealing with has `C:\Users\...`


ntamas:
Yup, `C:\Documents and Settings` is probably from the Windows XP / Windows 7 era. Thanks.

PR #591, opened 2022-10-10
Readthedocs

iosonofabio:
Trying out adaptations to build on RTD @ntamas 

Current status:
- [x] build set up
- [x] virtual environment config-ed
- [x] dependencies installed
- [x] C core builds
- [x] Python package builds
- [x] Python package installs into virtual env
- [x] Modularizing pure Python works
- [x] Sphinx starts
- [x] pydoctor finishes
- [x] sphinx finishes
- [x] output html files copied correctly to output folder
- [x] docs are live on RTD
- [x] get rid of Jekyll headers and set theme
- [x] polish appearance
- [x] tidy up code, esp. make sure distinct versions/branches/tags all have correct links
- [x] add RTD link checker
- [x] correct broken internal and external links
- [x] improve homepage
- [x] link older docs on homepage

https://readthedocs.org/projects/igraph/builds/

Here the output:

https://igraph.readthedocs.io/en/latest/


iosonofabio:
This now builds successfully on RTD after some hack due to race conditions between pydoctor and the RTD sphinx extension (sigh!). First successful build artifact here: https://igraph.readthedocs.io/en/latest/

Now I'll move on to improving appearance and tidying up code.

iosonofabio:
This is going quite well. Our docs are now deploying nicely to RTD including the API via pydoctor. Sphinx's linkcheck requires a recent version of `requests` to establish decent SSL handshakes with one (!) external link, so I'm trying to update that. In any case, it's not critical: the linkcheck works nicely on my machine.

doc/source/conf.py
```diff
@@ -30,11 +36,16 @@ def get_root_dir():
 
 def get_igraphdir():
     '''Get igraph folder'''
+    # Python version
     vmaj, vmin = sys.version_info[:2]
-    root_folder = get_root_dir()
+
+    if is_inside_rtd:
+        venv_folder = f'/home/docs/checkouts/readthedocs.org/user_builds/igraph/envs/{rtd_version}/'
```
ntamas, line 43: How stable is this path? Is this documented and guaranteed to look like this in readthedocs, or have you just inferred this?

iosonofabio: Reverse engineering... They don't seem to give guarantees and call all their executables via full path...

ntamas: Could we "detect" it somehow from inside `conf.py`, e.g., with `os.getcwd()` so it won't break if they decide to change the path later?

Or, maybe with inspecting `__file__` from inside `conf.py` and chopping off components from the end if needed?

iosonofabio: that's the thing: they keep the venvs separate from the source tree so that'd be equally messy I think. Ideally we don't need these paths at all and we can use `import igraph` to get what we need, however last I checked pydoctor was unhappy about it (a bug in pydoctor I believe). Happy to give it another go.

Otherwise we could import a dummy package (e.g. sphinx itself) and get its `__file__` and use that to reconstruct the venv location

ntamas: > we could import a dummy package (e.g. sphinx itself) and get its __file__ and use that to reconstruct the venv location

That sounds like a good idea! Let's try it and I'm happy to merge this PR afterwards if we managed to sort this out (or if we concluded that it's not worth it).

iosonofabio: `import importlib + <black magic> = 42`

elegant solution found

iosonofabio:
okay, I think this one's ready do merge AFAICT

szhorvat:
How does this work exactly? Does it regularly re-build the docs even for the dev version?

Now that it's merged, can it be connected to the official repo, or do we need to wait for the 0.10.2 release with that? I see now it's connect to @iosonofabio 's repo and has a big fat warning about not being official. https://readthedocs.org/projects/igraph/

I find it more pleasant to browse than the igraph.org ones though.

ntamas:
I've talked about this briefly with @iosonofabio after the end of today's dev meeting. He still wants to get the Sphinx Gallery stuff done so we can publish the examples with that. Once this is done, we can remove the warning.

As for the switch, there are two options:

1. We do the switch in parallel to 0.10.2 and then there's a clear situation: if you want the docs before 0.10.2, go to the archive on igraph.org, otherwise go to readthedocs.org.
2. We do the switch _after_ 0.10.1 but _before_ 0.10.2. In that case, people who want the _official_ 0.10.1 docs should still go to the homepage, but people searching via Google will find an "alternative" set of documentation on readthedocs.

I'd prefer option 1 as it causes less confusion.

iosonofabio:
As Tamas wrote, I'd be happier with option 1. I'll re-rig the RTD config against the correct repo etc. in the next 2-3 days.

Happy to hear you like it!

PR #593, opened 2022-10-14
Sphinx gallery

iosonofabio:
This PR aims at swapping our current custom sphinx extension to generate a gallery of examples with `sphinx-gallery` (https://sphinx-gallery.github.io/stable/index.html). I copied over all existing examples and improved many of them somewhat.

Main benefits:
- Adding an example is a matter of writing a Python file with a few additional rules re comments: no more messing with folders and subfolders
- Automatic links to source `.py` file of the example and a generated `.ipynb` Jupyter notebook file.
- One could host multiple sections in the gallery, e.g. networkx does it

Drawbacks/limitations:
- I couldn't get the animated GIF thumbnail to work yet, even though there are theoretically two ways to do it
- Raw scripts are outside `doc/source`, which can be a little confusing
- Fun fact: `sphinx-gallery` has experimental support for [binder](https://mybinder.org/), however that only works if the Jupyter notebooks are hosted on a GitHub repo. Because those files are generated on the fly by the extension on RTD, we cannot support binder yet. In fact the docs for `sphinx-gallery` itself gave up versioning in order to support binder, which is a funny choice I think.

Bottom line is that this should be merged in master AFAICT and then I'd like to give another pass to the entire docs next week to polish a few outstanding things. After that, the next big questions re python docs will be:

- how to handle re-export of modules (e.g. `igraph.drawing.plot` -> `igraph.plot`): pydoctor seems a little confused
- how to merge `objects.inv` from pydoctor and sphinx.
- how to create a separate gallery with longer "stories" (I've got a hunch already)

but those are all tales for another night...


iosonofabio:
I think we can merge this and then I'll switch the RTD config to use this repo

PR #599, opened 2022-10-18
docs: update Erdos-Renyi gallery example

szhorvat:
@Gomango999 Do you remember why you added the note at the end? I believe what is said there is not true, but if it is, that would be a bug, and we need to fix it.


Gomango999:
Ah I see, I genuinely thought it was true at the time. Some research has now shown that this is not the case. Feel free to delete it.

Issue #607, opened 2022-11-07
Support pandas "Int64" type


Adriankhl:
**What is the feature or improvement you would like to see?**
This produces an error:
```
import pandas as pd
import numpy as np
import igraph as ig

df = pd.DataFrame(np.array([[0, 1], [1, 1]]), columns=['a', 'b'])
df["a"] = df["a"].astype('Int64')
g = ig.Graph.DataFrame(df)
```

The error message:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In [19], line 1
----> 1 g = ig.Graph.DataFrame(df)

File ~/.pyenv/versions/3.11.0/lib/python3.11/site-packages/igraph/io/objects.py:424, in _construct_graph_from_dataframe(cls, edges, directed, vertices, use_vids)
    419 if use_vids:
    420     if not (
    421         str(edges.dtypes[0]).startswith("int")
    422         and str(edges.dtypes[1]).startswith("int")
    423     ):
--> 424         raise TypeError(
    425             f"Source and target IDs must be 0-based integers, found types {edges.dtypes.tolist()[:2]}"
    426         )
    427     elif (edges.iloc[:, :2] < 0).any(axis=None):
    428         raise ValueError("Source and target IDs must not be negative")

TypeError: Source and target IDs must be 0-based integers, found types [Int64Dtype(), dtype('int64')]
```

`Int64` is a nullable pandas int data type. I believe the type check is too strict. As long as the `Int64` does not contain any null value, the dataframe should be able to convert to an igraph object. The error message is also not precise for this case.

**Use cases for the feature**
Sometimes we just have to deal with a column with `Int64` as the data type, it adds a bit of annoyance if we have to convert it to `int64` every time.

**References**
https://pandas.pydata.org/docs/user_guide/integer_na.html

iosonofabio:
Thank you for reporting. I didn't know this type exists and have never used it in my 15 years of Python - hence the absence of support 😋

Let me look into it, sounds reasonable at face value but also not very high priority and there's an obvious workaround.

Can you open a PR? That'd speed things up

PR #609, opened 2022-11-08
Handle pandas nullable integer data type in Graph.DataFrame

Adriankhl:
Deal with https://github.com/igraph/python-igraph/issues/607.

Now the `Graph.DataFrame` can take columns with `Int64` data type, and it throws an error when there is any null value.

### DIFF
--- a/src/igraph/io/objects.py
+++ b/src/igraph/io/objects.py
@@ -417,15 +417,22 @@ def _construct_graph_from_dataframe(
         raise ValueError("The 'vertices' DataFrame must contain at least one column")
 
     if use_vids:
-        if not (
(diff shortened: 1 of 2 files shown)


iosonofabio:
Thank you for working the PR. I converted it to draft until it's ready for review.

It looks reasonable. Can you please add a test for this? Maybe one case that uses this type and fails with a null value and one case that passes?

Thank you

Adriankhl:
@iosonofabio Done

iosonofabio:
Let's see if it passes CI. If so I'll drag someone else for a final check and hopefully we can merge.

Thank you!

ntamas (approved):

ntamas:
Thank you!

PR #621, opened 2022-12-09
Add Python 3.11 classifier

gruebel:
Hi,

when I tried out `igraph` for one of our projects I was not sure, if you support Python 3.11, because the classifier was missing and the badge also didn't indicate it. Therefore I did some adjustments 🙂 

- Added Python 3.11 classifier to `setup.py`
- Adjusted the Python version version badge to atomically grab the supported versions from `pypi`, which are set via the classifiers
- Set Python version for 3.11 from RC to GA in GitHub workflow

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -294,7 +294,7 @@ jobs:
       - uses: actions/setup-python@v4
         name: Install Python
         with:
-          python-version: '3.11.0-rc.2'
+          python-version: '3.11'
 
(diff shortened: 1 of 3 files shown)


ntamas:
I haven't tested igraph on Python 3.11 yet so let's see what the CI tests have to say about it. If the tests pass, I'll be happy to merge this PR.

Thanks a lot!

ntamas (approved):

PR #631, opened 2023-01-23
wasm port effort, fix function pointers

pmp-p:
see https://blog.pyodide.org/posts/function-pointer-cast-handling/

1st test with PR applied  https://pmp-p.github.io/python-igraph-wasm/index.html?-i

### DIFF
--- a/src/_igraph/graphobject.c
+++ b/src/_igraph/graphobject.c
@@ -1620,7 +1620,7 @@ PyObject *igraphmodule_Graph_get_diameter(igraphmodule_GraphObject * self,
   } else {
     if (igraph_diameter(&self->g, 0,
           /* from, to, vertex_path, edge_path */
-          0, 0, &res, 0, 
+          0, 0, &res, 0,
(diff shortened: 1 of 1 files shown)


ntamas:
Hmm, nice catch, but shouldn't we get rid of the `_null` argument entirely then? `get_edgelist()` is defined as `METH_NOARGS` in the methods table, which means that it does not take any arguments (neither positional nor keyword). Other methods that are defined as `METH_NOARGS` have only the `self` argument.

pmp-p:
you really need the null to match exactly a METH_NOARGS signature

> Other methods that are defined as METH_NOARGS have only the self argument.

i did not had runtime error on them since i did not have time for running a testsuite, so i did not spot them, but they sure need to be fixed as well. for pygame i used an automated script to catch most of them https://gist.github.com/pmp-p/5641b7edfffba457d3eed875454ab5ca but as @dr0id mentionned, a parser is required

ntamas:
Thanks a lot! I'll fix the remaining `METH_NOARGS` methods in the code in a separate commit.

PR #635, opened 2023-01-29
split aarch64 build into manylinux and musllinux to prevent GHA timeout

gruebel:
@ntamas as suggested I split the `aarch64` build into 2 to prevent GHA timeout after 6 hours, which is currently quite tight and even timed out.


ntamas:
Thanks a lot!

PR #642, opened 2023-03-07
Add Spanish tutorial by @adriandiazlab

iosonofabio:
Similar to what was just merged into rigraph (thanks @ntamas!).

For now only one file with the long tutorial, plus some changes in `index.rst` to accomodate the link and TOC reference. This would already be incredibly useful for Spanish-native users.

Seems like remote `develop` is a little behind main so I branched off main instead. Happy to change if preferred.

Let me build locally and check everything's ok and then I'll mark ready for review. If anyone would like to give feedback already, please go ahead.


ntamas:
As I don't speak Spanish I can only say that the format seems OK. Does anyone want to give it a read first or shall I just merge it?

iosonofabio:
Let me give it a read
Fabio

iosonofabio:
I gave it a read and it mostly reads fine as far as my Spanish goes. Format was sometimes a little off, and functions references had been deleted once in a while.

Overall, if it passes CI I'd say we can merge. Much better than having no translation at all.

PR #663, opened 2023-05-04
Use graph layout attribute if found

### LINKED ISSUE #662: mpl plotting: layout as property of graph
Whenever a graph has a visual property set (e.g. vertex size) and the same property is specified in the arguments to `ig.plot`, the latter should take precedence. However, the logic seems garbled now and we make a mess.

This especially applies to:
- [ ] layouts

A quick check of the logic is what needs to be done here. Happy to do it over the next few days.

edit: actually I can reproduce this only with layout, so I made a PR specifically for that.

iosonofabio:
This partially fixes #662 based on what observed in the tutorial for the workshop.

If you do:

```python
import igraph as ig
 G = ig.Graph.Famous("Zachary")
G['layout'] = G.layout_fruchterman_reingold()
ig.plot(G)
```

The plotter does not know the graph already has a layout and will compute a new one. That means, if you plot multiple times you can get stochasticity if you don't reinitialize the RNG.

This PR introduces a few lines of code that check for a graph attribute "layout" and if present use it as a persistent layout.

Do you want me to write a test for this?


ntamas:
Yes, please add a test and then we can merge.

Issue #724, opened 2023-10-27
DAG Generation


SigireddyBalasai:
**What is the feature or improvement you would like to see?**
I need to create a random Direct Acyclic Graph For some statistical studies and i need to create using list of predefined nodes

**Use cases for the feature**
Explain when and for what purpose the feature would be useful.

**References**
List any relevant references (papers or books describing relevant algorithms).

ntamas:
You can generate a random undirected Erdos-Renyi graph with Graph.Erdos_Renyi() and then convert it into directed with the .to_directed(mode=“acyclic”) merhod. The returned graph will always be acyclic.

PR #725, opened 2023-10-27
Tutorial on DAG

iosonofabio:
Taking inspiration from #724, this adds a new tutorial on directed acyclic graphs.

I thought why not.

### DIFF
--- a/doc/examples_sphinx-gallery/generate_dag.py
+++ b/doc/examples_sphinx-gallery/generate_dag.py
@@ -1,0 +1,45 @@
+"""
+
+.. _tutorials-dag:
+
+======================
+Directed Acyclic Graph
+======================
+
+This example demonstrates how to create a random directed acyclic graph (DAG), which is useful in a number of contexts including for Git commit history.
+"""
+import igraph as ig
+import matplotlib.pyplot as plt
+import random
+
+
+# %%
+# First, we set a random seed for reproducibility
+random.seed(0)
+
+# %%
+# First, we generate a random undirected graph without loops
+g = ig.Graph.Erdos_Renyi(n=15, p=0.3, directed=False, loops=False)
+
+# %%
+# Then we convert it to a DAG *in place*
+g.to_directed(mode="acyclic")
+
+# %%
+# We can print out a summary of the DAG
+ig.summary(g)
+
+
+# %%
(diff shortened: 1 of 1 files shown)


ntamas (approved):

ntamas:
Thanks!

szhorvat:
Can you change this to use the G(n,m) model and add a comment that spells out that the method samples DAGs with the same number of vertices and edges _uniformly_?

szhorvat:
I'll push this change soon.

But a question: Does Sugiyama not do proper edge routing in Python, like it does in R? I see results like this:

image

I am expecting something like this instead:

image

ntamas:
> Does Sugiyama not do proper edge routing in Python, like it does in R? 

No it doesn't. More precisely: it can be called as `layout_sugiyama()`, in which case it returns the layout matrix but does not return the extended graph, for sake of compatibility with other `layout_...()` methods that also return the layout only, or it can be called as `layout_sugiyama(return_extended_graph=True)`, in which case it returns the extended graph with the dummy vertices _and_ the layout, but then you need to take care of setting up the plotting so that the additional vertices are hidden and only the edges are routed between them (and even then you won't have nice curved edges).

ntamas:
Not sure what the best approach would be when dealing with this in the tutorial. Something like this should do the trick:

```python
layout, extd_graph = g.layout("sugiyama", return_extended_graph=True)
ig.plot(
    extd_graph,
    vertex_size = [20] * g.vcount() + [0] * (extd_graph.vcount() - g.vcount())
    # the rest of arguments should be as they are in the current example
)
```

but then we'd still need to hide the arrowheads on all the edges that point to a dummy vertex, otherwise we end up with extra arrowheads. So it's a bit too much code for my taste for a simple tutorial example.
It shifts the focus from generating the DAG to plotting.

szhorvat:
In C/igraph 0.11 we should develop conventions for returning edge routing information. The extended graph approach might not be the best fit for other algorithms. Once we have this in place, there can be a convention `layout_` functions in Python as well. It would be nice to implement this for the new Python interface.

iosonofabio:
Indeed when we say "graph layout" casually we kind of mean "vertex layout", didn't think about that before.

I agree with @szhorvat that some form of support for routing would be nice to have after the refactor. That would also need a special infrastructure on the plotting side, to automate the issues highlighted above

Issue #726, opened 2023-10-27
Improvement of Readme.md


CrypticRevenger:
**What is the feature or improvement you would like to see?**

- Headers: Improved the headers to make them more descriptive and informative.

- Sentence Structure: Rewrote some sentences to enhance clarity and readability.

- Punctuation: Corrected punctuation and capitalization for consistency.

- Grammar: Fixed grammatical errors and awkward phrasings.

- Notes and Warnings: Clarified information under "Linking to an Existing igraph Installation" and added emphasis where necessary.

- Typographical Errors: Corrected typos and spelling errors.

szhorvat:
Are you trying to open a pull request?

CrypticRevenger:
> Are you trying to open a pull request?

yes

CrypticRevenger:
@szhorvat can you please merge it. 
It will be very helpful to me. 
Thank you.

szhorvat:
> > Are you trying to open a pull request?
> 
> yes

In that case there is no need to open an issue. Please open issues only for bugs, or well-motivated feature requests.

> It will be very helpful to me.

What personal benefit do you gain from this PR getting merged? I hope this is not about improving [your contribution stats](https://github.com/CrypticRevenger#-github-stats). Please also consider whether the changes you propose are a clear benefit to the project.

CrypticRevenger:
> What personal benefit do you gain from this PR getting merged? I hope this is not about improving [your contribution stats](https://github.com/CrypticRevenger#-github-stats). Please also consider whether the changes you propose are a clear benefit to the project.
I am trying to contribute in hacksquad and I am totally new to git. That's why  I am trying to contribute in different repos.

CrypticRevenger:
@szhorvat Can you please suggest me what can I do in this repo to make it better.

szhorvat:
I made a suggestion about how to contribute with writing-based tasks within the PR you opened. Please see the comment there.

PR #727, opened 2023-10-27
Improvement of Readme.md #726

CrypticRevenger:
PR no :- #726

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,102 +6,95 @@
 
 # Python interface for the igraph library
 
-igraph is a library for creating and manipulating graphs.
-It is intended to be as powerful (ie. fast) as possible to enable the
-analysis of large graphs.
+igraph is a library designed for creating and manipulating graphs. Its primary aim is to be as powerful and fast as possible, making it suitable for analyzing large graphs.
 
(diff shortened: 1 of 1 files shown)


CrypticRevenger:
@ntamas Can you please review my PR.
Thank you.

iosonofabio:
Apologies if this comes across a little rude, but these changes look like ChatGPT. Did you use it for this @CrypticRevenger?

szhorvat:
From my perspective, it isn't a problem if it was or wasn't done with ChatGPT ... The issue here is that most of these changes simply rephrase sentences without improving them. There are a large number of changes and it's difficult to go through all of them. I do not agree with some of the new phrasings. I do not agree igraph's "**primary aim** is to be as powerful and fast as possible".

I do not make the decision about merging this, but my recommendation is not to.

My suggestion to you is to identify specific problems, make it clear what they are, and fix _only_ those problems. There are many, many functions whose documentation needs substantial improvements. If you'd like to contribute by writing, I suggest starting with the docs of specific functions, make sure the description is _clear_ and _concise_ and up to date with the docs of [the underlying C function](https://igraph.org/c/html/master/).

PR #763, opened 2024-03-06
Update from deprecated PyCObject to PyCapsule

DavidRConnell:
I was trying to use python-igraph's api and got compiler warnings for implicit declaration of `PyCObject_Check` and `PyCObject_AsVoidPtr`. The [CObject api has been deprecated](https://python.readthedocs.io/en/v2.7.2/c-api/cobject.html). It looks like this got updated elsewhere but missed in `igraphmodule_api.h`. Pretty much just copied from the [Providing a C API for an Extension Module](https://python.readthedocs.io/en/stable/extending/extending.html#using-capsules) example. Changing to my branch fixes the previous errors I was getting with the missing `PyCObject_*` symbols.

### DIFF
--- a/src/_igraph/igraphmodule_api.h
+++ b/src/_igraph/igraphmodule_api.h
@@ -56,25 +56,8 @@ extern "C" {
 
   /* Return -1 and set exception on error, 0 on success */
   static int import_igraph(void) {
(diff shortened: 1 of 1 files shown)


ntamas:
Thank you!

PR #767, opened 2024-03-20
Bug fix in libraries.py

rmmaf:
The 'pop' function in line 58 is called wrongly (with '[]' instead of '()'), resulting in:

TypeError: 'builtin_function_or_method' object is not subscriptable


ntamas (approved):
Thanks, nice catch!

PR #770, opened 2024-04-20
Linking with existing igraph on msys2

### LINKED ISSUE #769: Building with an external igraph library on Windows
quote from https://github.com/igraph/python-igraph/blob/19b5a45d0e987397256efa48fd447cec5e7315c9/README.md?plain=1#L161
Well pkg-config might not be supported, but pkgconf is https://packages.msys2.org/package/mingw-w64-x86_64-pkgconf which shims pkg-config... so this is not true. You can link to dll libraries lol in Windows

Kreijstal:
Expanded the instructions on how to build on msys2 and link to the existing igraph installation.

I've been thinking for a while if the msys2/mingw distribution should provide a package for python-igraph itself, but these are instructions to build from source, so it should suffice, shouldn't also other distros do this?
fixes #769

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,8 @@ where `[arch]` is either `Win32` for 32-bit builds or `x64` for 64-bit builds.
 Also, when building in MSYS2, you need to set the `SETUPTOOLS_USE_DISTUTILS`
 environment variable to `stdlib`; this is because MSYS2 uses a patched version
 of `distutils` that conflicts with `setuptools >= 60.0`.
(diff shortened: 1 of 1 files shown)


szhorvat:
> I've been thinking for a while if the msys2/mingw distribution should provide a package for python-igraph itself

If you can contribute such a package to MSYS2, that would be nice.

ntamas (approved):

ntamas:
Thanks a lot!

PR #786, opened 2024-06-28
feat: `is_clique()` and `is_independent_vertex_set()`

szhorvat:
One question is whether we should drop `is_complete()` from Python now that `g.is_clique()` (with no parameters) performs the same function, and dispatches to the same C function.

Perhaps just keep it for discoverability?

### DIFF
--- a/src/_igraph/graphobject.c
+++ b/src/_igraph/graphobject.c
@@ -568,6 +568,77 @@ PyObject *igraphmodule_Graph_is_complete(igraphmodule_GraphObject* self, PyObjec
 
 
 /** \ingroup python_interface_graph
+ * \brief Checks whether a given vertex set forms a clique
+ */
+PyObject *igraphmodule_Graph_is_clique(igraphmodule_GraphObject * self,
+                                       PyObject * args, PyObject * kwds)
+{
+  PyObject *list = Py_None;
+  PyObject *directed = Py_False;
+  igraph_bool_t res;
+  igraph_vs_t vs;
+
+  static char *kwlist[] = { "vertices", "directed", NULL };
+
+  if (!PyArg_ParseTupleAndKeywords(args, kwds, "|OO", kwlist, &list, &directed)) {
+    return NULL;
+  }
+
+  if (igraphmodule_PyObject_to_vs_t(list, &vs, &self->g, NULL, NULL)) {
+    return NULL;
+  }
+
+  if (igraph_is_clique(&self->g, vs, PyObject_IsTrue(directed), &res)) {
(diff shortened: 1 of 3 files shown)


szhorvat:
Resolved conflicts and rebased.

ntamas:
> Perhaps just keep it for discoverability?

Yes.

src/_igraph/graphobject.c
```diff
@@ -13728,6 +13799,22 @@ struct PyMethodDef igraphmodule_Graph_methods[] = {
    "@return: C{True} if it is complete, C{False} otherwise.\n"
    "@rtype: boolean"},
 
+  {"is_clique", (PyCFunction) igraphmodule_Graph_is_clique,
+   METH_VARARGS | METH_KEYWORDS,
+   "is_clique(vertices, directed=False)\n--\n\n"
```
ntamas, line 13804: `is_clique()` supports `vertices=None` and `vertices` itself is optional. Here it seems that `vertices` is mandatory. I don't mind either way, you can make it mandatory or you can leave it as-is (when it supports `None`), just make sure that the docs are consistent with the behaviour.

src/_igraph/graphobject.c
```diff
@@ -13728,6 +13799,22 @@ struct PyMethodDef igraphmodule_Graph_methods[] = {
...
+   "Decides whether a set of vertices is a clique, i.e. a fully connected subgraph.\n\n"
+   "@param vertices: a list of vertex IDs\n"
+   "@param directed: whether to require mutual connections between vertex pairs\n"
+   "    in directed graphs.\n"
+   "@return: C{True} is the given vertex set is a clique, C{False} if not.\n"},
+
+  {"is_independent_vertex_set", (PyCFunction) igraphmodule_Graph_is_independent_vertex_set,
+   METH_VARARGS | METH_KEYWORDS,
+   "is_independent_vertex_set(vertices)\n--\n\n"
```
ntamas, line 13813: Same as above: `vertices=None` is supported now and it is optional. The docs should be consistent with the behaviour.

ntamas (changes requested):
Feel free to merge after addressing the review comments.

src/_igraph/graphobject.c
szhorvat: Fixed.

src/_igraph/graphobject.c
szhorvat: Fixed.

szhorvat:
@ntamas PR merged, I think we're good for a release.

PR #787, opened 2024-07-03
refactor: rewire() now does 10 times the number of edges trials by default

### LINKED ISSUE #775: The default number of trials in `rewire()` should be proportional to the edge count
The default number of edge switching trials in rewire() should be proportional to the edge count. Currently this is fixed at 1000.

This is important, since we're trying to lead inexperienced users on the right track.

It led to a major misunderstanding in [this recent preprint](https://arxiv.org/abs/2405.05400), see footnote `*` in Table 2.  I just emailed the authors.

As a start I suggest using 10 times the edge count until we get arguments for a better factor than 10. The preprint I linked mentions Milo et al recommending 100, but I wouldn't put much weight on that: it's very old and ultimately it's a response to a criticism of their original biased randomization method, not actually focusing on edge switching methods.

szhorvat:
Closes #775 

@ntamas, can you help with this? What should I specify as the default for `n`? I'm pretty sure that the way I wrote `n=None` is not correct.

### DIFF
--- a/src/_igraph/graphobject.c
+++ b/src/_igraph/graphobject.c
@@ -6432,14 +6432,19 @@ PyObject *igraphmodule_Graph_rewire(igraphmodule_GraphObject * self,
                                     PyObject * args, PyObject * kwds)
 {
   static char *kwlist[] = { "n", "mode", NULL };
-  Py_ssize_t n = 1000;
-  PyObject *mode_o = Py_None;
+  PyObject *n_o = Py_None, *mode_o = Py_None;
+  igraph_integer_t n = 10 * igraph_ecount(&self->g); /* TODO overflow check */
   igraph_rewiring_t mode = IGRAPH_REWIRING_SIMPLE;
 
-  if (!PyArg_ParseTupleAndKeywords(args, kwds, "|nO", kwlist, &n, &mode_o))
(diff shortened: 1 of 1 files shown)


ntamas:
`PyArg_ParseTupleAndKeywords(args, kwds, "|nO", kwlist, &n, &mode_o)` -- this function parses `n` as a Python integer `int` and saves it in a `Py_ssize_t`, that's why you cannot simply pass `None` for `n` right now (see [here](https://docs.python.org/3/c-api/arg.html#parsing-arguments) for more details). If you wanted to support `None`, you would need to change the format string to `|OO`, parse `n` into a `PyObject*` first, check whether it's `None` (with `== Py_None`) and if not, try to cast it into a `Py_ssize_t` with `PyNumber_AsSsize_t` (see [here](https://docs.python.org/3/c-api/number.html#c.PyNumber_AsSsize_t) ).

szhorvat:
What I want is that the effective default be `1000*ecount`. What is the best way to do this? Should the formal default be `None`? If yes, I now understand how to do that and can update this PR. Let me know.

ntamas:
Yes, that's the way to go, that's how I would do it in pure Python as well.

szhorvat:
@ntamas Have another look, I used `igraphmodule_PyObject_to_integer_t()` instead of `PyNumber_AsSsize_t`. Is this okay?

ntamas (approved):

PR #806, opened 2024-11-13
Implement simple cycle search Python binding

GenieTim:
Closes #804; here, the output is determined by the corresponding parameter.

### DIFF
--- a/src/_igraph/graphobject.c
+++ b/src/_igraph/graphobject.c
@@ -7813,6 +7813,65 @@ PyObject *igraphmodule_Graph_minimum_cycle_basis(
   return result_o;
 }
 
+
+PyObject *igraphmodule_Graph_simple_cycles(
+  igraphmodule_GraphObject *self, PyObject *args, PyObject *kwds
+) {
+  PyObject *mode_o = Py_None;
+  PyObject *output_o = Py_None;
+  PyObject *min_cycle_length_o = Py_None;
+  PyObject *max_cycle_length_o = Py_None;
+
+  // argument defaults: no cycle limits
+  igraph_integer_t mode = IGRAPH_OUT;
+  igraph_integer_t min_cycle_length = -1;
+  igraph_integer_t max_cycle_length = -1;
+  igraph_bool_t use_edges = false;
+
+  static char *kwlist[] = { "mode", "min_cycle_length", "max_cycle_length", "output" NULL };
+
+  if (!PyArg_ParseTupleAndKeywords(args, kwds, "|OOOO", kwlist, &mode_o, &min_cycle_length_o, &max_cycle_length_o, &output_o))
+    return NULL;
+
+  if (mode_o != Py_None && igraphmodule_PyObject_to_integer_t(mode_o, &mode))
+    return NULL;
+
+  if (min_cycle_length_o != Py_None && igraphmodule_PyObject_to_integer_t(min_cycle_length_o, &min_cycle_length))
+    return NULL;
+
+  if (max_cycle_length_o != Py_None && igraphmodule_PyObject_to_integer_t(max_cycle_length_o, &max_cycle_length))
+    return NULL;
+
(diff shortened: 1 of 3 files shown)


szhorvat:
I would suggest using simply `min` and `max` for the parameter names. This will be consistent with `cliques()`, https://python.igraph.org/en/stable/api/igraph.GraphBase.html#cliques, which is another function with lower and upper limits on the objects returned.

src/_igraph/graphobject.c
```diff
@@ -16563,6 +16622,24 @@ struct PyMethodDef igraphmodule_Graph_methods[] = {
    "  no guarantees are given about the ordering of edge IDs within cycles.\n"
    "@return: the cycle basis as a list of tuples containing edge IDs"
   },
+  {"simple_cycles", (PyCFunction) igraphmodule_Graph_simple_cycles,
+   METH_VARARGS | METH_KEYWORDS,
+   "simple_cycles(mode=None, min=0, max=-1, output=\"vpath\")\n--\n\n"
```
szhorvat, line 16627:
```suggestion
   "simple_cycles(mode=None, min=-1, max=-1, output=\"vpath\")\n--\n\n"
```

Of course 0 works too, but using `-1` would make it easier to create consistent and predictable interfaces in the future.

szhorvat:
@ntamas Could you please have a look at this when you get the time? This is very valuable functionality.

src/_igraph/graphobject.c
```diff
@@ -7813,6 +7813,65 @@ PyObject *igraphmodule_Graph_minimum_cycle_basis(
...
+    return NULL;
+
+  if (max_cycle_length_o != Py_None && igraphmodule_PyObject_to_integer_t(max_cycle_length_o, &max_cycle_length))
+    return NULL;
+
+  if (igraphmodule_PyObject_to_vpath_or_epath(output_o, &use_edges))
+    return NULL;
+
+  igraph_vector_int_list_t vertices;
+  igraph_vector_int_list_init(&vertices, 0);
```
ntamas, line 7849: Even though we are allocating a vector of size 0 here, the allocation may still fail (we never really allocate a buffer of zero bytes, there's a minimum amount of memory that we allocate even for an empty vector). Therefore, this call needs to be guarded against failures with the usual pattern:

```
if (igraph_vector_int_list_init(&vertices, 0)) {
    igraphmodule_handle_igraph_error();
    return NULL;
}
```

Same for the allocation of the edge vector below.

ntamas (changes requested):

src/_igraph/graphobject.c
GenieTim: Thanks, good point, I introduced these two guards in my latest commits

src/_igraph/graphobject.c
```diff
@@ -7813,6 +7813,71 @@ PyObject *igraphmodule_Graph_minimum_cycle_basis(
...
+    return NULL;
+
+  igraph_vector_int_list_t vertices;
+  if (igraph_vector_int_list_init(&vertices, 0)) {
+    igraphmodule_handle_igraph_error();
+    return NULL;
+  }
+  igraph_vector_int_list_t edges;
+  if (igraph_vector_int_list_init(&edges, 0)) {
+    igraphmodule_handle_igraph_error();
```
szhorvat, line 7855:
```suggestion
    igraph_vector_int_list_destroy(&vertices);
    igraphmodule_handle_igraph_error();
```

szhorvat: If this fails, we also need to destroy the previously allocated `vertices`.

ntamas:
> @ntamas Do you have any other concerns about this one (other than the comment I made above)?

Nope, no other concerns, but what you pointed out needs to be addressed.

szhorvat:
Shall we merge this?

szhorvat (approved):

PR #810, opened 2024-12-22
setup.py: import bdist_wheel from setuptools

m1-s:
This PR removes the direct dependency from `wheel` and imports `bdist_wheel` from setuptools instead. This change is motivated by https://github.com/pypa/setuptools/pull/4647. 

Other libs have done similar changes as this PR https://github.com/yaml/pyyaml/pull/823/files.

Disclaimer: I am not a python developer. I am just trying to fix a piece of software that depends on igraph and broke when being built in an environment with more recent dependency versions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,5 @@
 [build-system]
 requires = [
-    "wheel",
     # pin setuptools:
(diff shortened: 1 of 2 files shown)


ntamas:
CI failure above is most likely unrelated to this change so I'll just merge and try to fix the CI failure in `main`. Thanks a lot!

Issue #811, opened 2025-01-21
Question re `cutoff` kwarg on harmonic_centrality()


aothms:
**Describe the bug**

I cannot seem to understand how the cutoff parameter works on harmonic_centrality() and would be tempted to think it's wrong, because it returns:

- the same result for any value passed to it 
- doesn't approach the value returned without cutoff for a very high cutoff

**To reproduce**

```
>>> r100 = Graph.Ring(n=100, circular=True)
>>> set(r100.harmonic_centrality())
{0.09069101693594803}
>>> set(r100.harmonic_centrality(cutoff=1))
{0.020202020202020204}
>>> set(r100.harmonic_centrality(cutoff=10))   
{0.020202020202020204}
>>> set(r100.harmonic_centrality(cutoff=1e100))
{0.020202020202020204}
```

**Version information**

```
>>> igraph.version.__version__
'0.11.8'
```

aothms:
Yes, thanks for that, based on that I gathered the courage to have a look:

It seems the arguments are not in the correct order.

C: first `normalized` then `cutoff`

```c
IGRAPH_EXPORT igraph_error_t igraph_harmonic_centrality_cutoff(const igraph_t *graph, igraph_vector_t *res,
                                                    const igraph_vs_t vids, igraph_neimode_t mode,
                                                    const igraph_vector_t *weights,
                                                    igraph_bool_t normalized,
                                                    igraph_real_t cutoff);
```

https://github.com/igraph/igraph/blob/35dde06c97eeec212a0a613fade6768be020a041/include/igraph_centrality.h#L56-L60

But what gets passed from python: first `cutoff` then `normalized`:

```c
    if (igraph_harmonic_centrality_cutoff(&self->g, &res, vs, mode, weights,
        PyFloat_AsDouble(cutoff_num), PyObject_IsTrue(normalized_o))) {
```

https://github.com/igraph/python-igraph/blob/8d2e9bc056c9d27ae859b1716cf9e8916e8e06e6/src/_igraph/graphobject.c#L4914-L4915

I'll send in a PR

PR #812, opened 2025-01-21
Update graphobject.c - reverse cutoff, normalized args

aothms:
See #811 for context.

### DIFF
--- a/src/_igraph/graphobject.c
+++ b/src/_igraph/graphobject.c
@@ -4832,7 +4832,7 @@ PyObject *igraphmodule_Graph_closeness(igraphmodule_GraphObject * self,
       return NULL;
     }
     if (igraph_closeness_cutoff(&self->g, &res, NULL, NULL, vs, mode, weights,
-        PyFloat_AsDouble(cutoff_num), PyObject_IsTrue(normalized_o))) {
+        PyObject_IsTrue(normalized_o), PyFloat_AsDouble(cutoff_num))) {
       igraph_vs_destroy(&vs);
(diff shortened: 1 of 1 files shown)


szhorvat:
Thanks for this! I had a brief look when you opened an issue, but I wasn't able to see what's wrong just by reading the code. It's indeed the argument order.

szhorvat (approved):

src/_igraph/graphobject.c
```diff
@@ -4832,7 +4832,7 @@ PyObject *igraphmodule_Graph_closeness(igraphmodule_GraphObject * self,
       return NULL;
     }
     if (igraph_closeness_cutoff(&self->g, &res, NULL, NULL, vs, mode, weights,
-        PyObject_IsTrue(normalized_o), PyFloat_AsDouble(cutoff_num))) {
+        PyFloat_AsDouble(cutoff_num), PyObject_IsTrue(normalized_o))) {
```
ntamas, line 4835: The original order seems to be correct here (cf. `include/igraph_centrality.h` in the igraph/igraph repo) so I think this is not needed here.

ntamas (changes requested):

src/_igraph/graphobject.c
szhorvat: Sorry, I missed this when reviewing. You're right.

ntamas (approved):

ntamas:
Thanks! I'll merge this upon CI pass.

PR #818, opened 2025-03-22
Added Interface for Netzschleuder .gml.zst files

AdityaChaubeyIITM:
I have added a new file repositories.py under src/igraph/io/ , for the creation of graph from netzschleuder.
It lets the user just specify the name of the dataset and get the single network in the dataset as a result if there is only one network in the dataset, otherwise it asks for the sub-network's name as well.
It temporarily downloads the .gml.zst file, decompresses it and then creates a graph out of the data. The graphs scope is limited to the data presentation in .gml files in netzschleuder. 
This is supposed to be an attempt to open discussion on methods to import netzschleuder dataset via '.gml' datatype files. It can be further modified to take care of the issues arising due to incorrect handling of edge attributes in .gml files.


szhorvat:
Just a heads up that commit b458a31a2f1885ffbfaa1fd2ddb97b4e9be2b158 broke some tests. You don't need to worry about these, I'll look into sorting them out.

AdityaChaubeyIITM:
image
thank you @szhorvat !
However for this particular case Build test says it broke due to missing zstandard module, I guess that can be fixed by adding zstandard to the requirements.txt.
Please let me know if that is what is needed, I shall correct it from my end then.

Similar issue happened earlier with the previous one [adaeae0](https://github.com/igraph/python-igraph/pull/818/commits/adaeae073c60925bd1587a5ef36c1f734d06d390)  also broke because of missing requests.

AdityaChaubeyIITM:
Greetings @szhorvat ! can you please review this commit? , there seems to be something missing in the way i am defining the function imports.

PR #820, opened 2025-03-26
Added Stochastic Variability in Community Detection Algorithms

SKG24:
Used Chatgpt for understanding the functions and theory around the mathematical process.

At first run:
Screenshot 2025-03-26 at 9 25 41 PM

At second run:
Screenshot 2025-03-26 at 9 27 55 PM

This supports stochastic method might give wildly different answers on a network without any significant community structure, while it gives consistent answers on one that has obvious communities.

Since the number of iterations is set to 50, even in structured graphs, slight variations in similarity scores may be observed.


szhorvat:
I won't have time to look in detail today, but I checked whether the docs build with this change, and unfortunately they do not. Can you please check if you can fix this? You can build the docs using `scripts/mkdoc.sh -c`.

SKG24:
Screenshot 2025-03-27 at 9 56 39 AM

Issue Description:
The `script mkdoc.sh -c` fails with the following error:

*AttributeError:* module 'igraph' has no attribute '_igraph'. Did you mean: 'Graph'?
This issue is persistent even after:

- Reinstalling python-igraph
- Passing clustering objects instead of membership lists in compare_communities()
- Ensuring the same code works in Google Colab

ntamas:
I think I've fixed the build issue in the `main` branch; please try again and let me know if it still doesn't work.

SKG24:
Thank you! It worked. 

https://github.com/user-attachments/assets/6d983690-110c-4822-930f-910a89554453

doc/examples_sphinx-gallery/stochastic_variability.py
```diff
@@ -0,0 +1,93 @@
+"""
+.. _tutorials-stochastic-variability:
+
+=========================================================
+Stochastic Variability in Community Detection Algorithms
+=========================================================
+
+This example demonstrates the variability of stochastic community detection methods by analyzing the consistency of multiple partitions using similarity measures (NMI, VI, RI) on both random and structured graphs.
```
szhorvat, line 8: Please spell out the names of similarity measures. If you like, you can add the abbreviations in parentheses.

doc/examples_sphinx-gallery/stochastic_variability.py
```diff
@@ -0,0 +1,93 @@
...
+
+"""
+# %%
+# Import Libraries
+import igraph as ig
+import numpy as np
```
szhorvat, line 14: Do not import libraries you don't use.

doc/examples_sphinx-gallery/stochastic_variability.py
```diff
@@ -0,0 +1,93 @@
...
+
+=========================================================
+Stochastic Variability in Community Detection Algorithms
+=========================================================
+
+This example demonstrates the variability of stochastic community detection methods by analyzing the consistency of multiple partitions using similarity measures (NMI, VI, RI) on both random and structured graphs.
+
+"""
+# %%
+# Import Libraries
```
szhorvat, line 12: Please do not capitalize words without good reason.

```suggestion
# Import libraries:
```

doc/examples_sphinx-gallery/stochastic_variability.py
```diff
@@ -0,0 +1,93 @@
...
+import igraph as ig
+import numpy as np
+import matplotlib.pyplot as plt
+import itertools
+
+# %%
+# First, we generate a graph.
+# Generates a random Erdos-Renyi graph (no clear community structure)
+def generate_random_graph(n, p):
+    return ig.Graph.Erdos_Renyi(n=n, p=p)
```
szhorvat, line 22: Do not omit diacritics. It is Erdős-Rényi.

For clarity, do indicate that it is an Erdős-Rényi $G(n,p)$ graph (i.e. not $G(n,m)$).

Do we really need to define new functions to generate these graphs? This function just wraps `Graph.Erdos_Renyi`.

doc/examples_sphinx-gallery/stochastic_variability.py
```diff
@@ -0,0 +1,93 @@
...
+  
+# %%
+# Generates a clustered graph with clear communities using the Stochastic Block Model (SBM)
+def generate_clustered_graph(n, clusters, intra_p, inter_p):
+    block_sizes = [n // clusters] * clusters
+    prob_matrix = [[intra_p if i == j else inter_p for j in range(clusters)] for i in range(clusters)]
+    return ig.Graph.SBM(sum(block_sizes), prob_matrix, block_sizes)
```
szhorvat, line 29: Could we simplify code the code while also make it more illustrative and use empirical network data here?

You could try the karate club network, the Les Miserables network (already available in the same directory) or perhaps the famous [Jazz musicians network](https://networks.skewed.de/net/jazz_collab). See which one gives a nicer result.

For the random graph, let's use one that has the same vertex count and density as the empirical one. Measure the density and pass it as the $p$ parameter of the $G(n,p)$ model. Alternatively, measure the edge count and pass it as the $m$ parameter of the $G(n,m)$ model.

doc/examples_sphinx-gallery/stochastic_variability.py
```diff
@@ -0,0 +1,93 @@
...
+def compute_pairwise_similarity(partitions, method):
+    """Computes pairwise similarity measure between partitions."""
+    scores = []
+    for p1, p2 in itertools.combinations(partitions, 2):
+        scores.append(ig.compare_communities(p1, p2, method=method))
+    return scores
+
+# %%
+# Stochastic Community Detection
+# Runs Louvain's method iteratively to generate partitions
```
szhorvat, line 42: This is called the [Louvain](https://en.wikipedia.org/wiki/UCLouvain) method, not Louvain's method.

Can you include a short explanation of why the result is different on each run? This is often a point of confusion for empirical researchers who are inexperienced in data analysis.

This is a modularity maximization method. Since the exact maximization of modularity is NP-hard, the Louvain method uses a greedy heuristic, processing vertices in a random order.

doc/examples_sphinx-gallery/stochastic_variability.py
```diff
@@ -0,0 +1,93 @@
...
+
+for i, (random_scores, clustered_scores, measure) in enumerate(measures):
+    axes[i][0].hist(random_scores, bins=20, alpha=0.7, color=colors[i], edgecolor="black")
+    axes[i][0].set_title(f"Histogram of {measure} - Random Graph")
+    axes[i][0].set_xlabel(f"{measure} Score")
+    axes[i][0].set_ylabel("Frequency")
+    
+    axes[i][1].hist(clustered_scores, bins=20, alpha=0.7, color=colors[i], edgecolor="black")
+    axes[i][1].set_title(f"Histogram of {measure} - Clustered Graph")
+    axes[i][1].set_xlabel(f"{measure} Score")
```
szhorvat, line 84: Could you please plot the probability density instead of counts? While doesn't make a difference here, it is generally good practice, and it becomes relevant when comparing datasets of different sizes.

doc/examples_sphinx-gallery/stochastic_variability.py
```diff
@@ -0,0 +1,93 @@
...
+
+plt.tight_layout()
+plt.show()
+
+# %%
+# The results are plotted as histograms for random vs. clustered graphs, highlighting differences in detected community structures.
+#The key reason for the inconsistency in random graphs and higher consistency in structured graphs is due to community structure strength:
+#Random Graphs: Lack clear communities, leading to unstable partitions. Stochastic algorithms detect different structures across runs, resulting in low NMI, high VI, and inconsistent RI.
+#Structured Graphs: Have well-defined communities, so detected partitions are more stable across multiple runs, leading to high NMI, low VI, and stable RI.
```
szhorvat, line 93: Could you please be explicit about the range and interpretation of the three measures? NMI and RI are in $[0,1]$, and larger values indicate higher similarity. VI is a distance metric, thus lower values indicate higher similarity.

szhorvat (changes requested):
This is a nice illustration!

I left a few comments for improvement.

Please remove changes to the `sg_execution_times.rst` file. This file was likely committed by accident, and I think we should remove it (but not as part of this PR).

doc/examples_sphinx-gallery/stochastic_variability.py
szhorvat: Also, please adjust the NMI and RI histograms to span the range $[0,1]$, and adjust the VI histogram to have a lower bound of 0.

SKG24:
I have made the suggested changes. 

https://github.com/user-attachments/assets/bb7f31e7-debd-4b40-b63b-a0d233d9272e

SKG24:
> This is a nice illustration!
> 
> I left a few comments for improvement.
> 
> Please remove changes to the `sg_execution_times.rst` file. This file was likely committed by accident, and I think we should remove it (but not as part of this PR).

Should I directly delete this file from PR through files changed section?

szhorvat:
> Should I directly delete this file from PR through files changed section?

Yes, that would be fine.

szhorvat:
Continued in #840. I can't push to this as the work was done on the `main` branch instead of a separate branch.

Issue #819, opened 2025-03-22
Create more Galery examples


szhorvat:
The python-igraph documentation includes a gallery of example applications:

https://python.igraph.org/en/stable/tutorials/index.html

This issue is to implement more examples.  If you have ideas for examples, post them here.  For inspiration, you may also look at [the IGraph/M documentation](http://szhorvat.net/mathematica/IGDocumentation/), which includes many visual examples.

 - [ ] Clique percolation.
 - [ ] Personalized PageRank on a grid (see http://szhorvat.net/mathematica/IGDocumentation/#pagerank)
 - [ ] Motif analysis with nice plots. Compare 3-motif counts in some directed network with a degree-constrained null model.
 - [ ] _... your idea here! ..._

SKG24:
I have prepared a documentation file explaining Partition Similarity Measure, including code snippets in Python. Could you confirm whether the appropriate path for adding examples is doc/examples_sphinx-gallery/? I want to ensure the file is placed correctly within the documentation structure.

https://github.com/SKG24/igraph_learning/blob/main/Community%20Detection%20and%20Partition%20Similarity%20using%20igraph.md

BeaMarton13:
Hi! I'm planning to work on the _Personalized PageRank on a grid_ topic. Does that sound good?

SKG24:
Hey! should I, continue adding the Partition Similarity Measure example with a concise and focused explanation, ensuring it aligns with the gallery’s principles—brief, to the point, and accompanied by an attractive visualization.

Can I proceed with raising a PR under the gallery section?

szhorvat:
@SKG24 It seems you are detecting communities with multiple methods then comparing the results using various metrics. I feel that while this will be useful for the future tutorial, it does not make an ideal example for the gallery. It does not show something that a newcomer might find difficult to figure out (other than visualizing a community structure, which already has an example), but it does bring many questions which can't easily be answered in the confines of such an example. Why do different methods give different results? Which is the right one? Why does even a single method sometimes give different results across multiple runs?

Ideally, a gallery example should try to either demonstrate a task that's popular but it's not obvious for a newcomer how to do it, or it should make a single point.  If you are interested in using partition similarity measures, you could for example demonstrate how a stochastic method might give wildly different answers on a network without any significant community structure, while it gives consistent answers on one that has obvious communities. For example, use a stochastic method, generate multiple clusterings, then plot a histogram of pairwise similarities between these.

BeaMarton13:
[...]

PR #821, opened 2025-03-26
docs: add personalized PageRank on a grid tutorial

BeaMarton13:
This PR implements a tutorial for calculating and visualizing personalized PageRank on a grid, as suggested [here](https://github.com/igraph/python-igraph/issues/819).

### DIFF
--- a/doc/examples_sphinx-gallery/personalized_pagerank.py
+++ b/doc/examples_sphinx-gallery/personalized_pagerank.py
@@ -1,0 +1,81 @@
+"""
+.. _tutorials-personalized_pagerank:
+
+===============================
+Personalized PageRank on a grid
+===============================
+
+This example demonstrates how to calculate and visualize personalized PageRank on a grid. We use the :meth:`igraph.GraphBase.personalized_pagerank` method, and demonstrate the effects on a grid graph.
+"""
+
+import matplotlib.cm as cm
+import matplotlib.pyplot as plt
+import numpy as np
+
+import igraph as ig
+
+
+# %%
+# We define a function that plots the graph on a Matplotlib axis, along with
+# its personalized PageRank values. The function also generates a
+# color bar on the side to see how the values change.
+# We use `Matplotlib's Normalize class <https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.Normalize.html>`_
+# to set the colors and ensure that our color bar range is correct.
+def plot_pagerank(graph: ig.Graph, p_pagerank: list[float]):
+    """Plots personalized PageRank values on a grid graph with a colorbar.
+
+    Parameters
+    ----------
+    graph : ig.Graph
+        graph to plot
+    p_pagerank : list[float]
+        calculated personalized PageRank values
+    """
+    # Create the axis for matplotlib
+    _, ax = plt.subplots(figsize=(8, 8))
+
+    # Create a matplotlib colormap
+    # coolwarm goes from blue (lowest value) to red (highest value)
+    cmap = cm.coolwarm
+
+    # Normalize the PageRank values for colormap
+    normalized_pagerank = ig.rescale(p_pagerank)
+
+    graph.vs["color"] = [cmap(pr) for pr in normalized_pagerank]
+    graph.vs["size"] = ig.rescale(p_pagerank, (20, 40))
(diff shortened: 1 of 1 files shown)


ntamas (approved):

doc/examples_sphinx-gallery/personalized_pagerank.py
```diff
@@ -0,0 +1,81 @@
+"""
+.. _tutorials-personalized_pagerank:
+
+===============================
+Personalized PageRank on a grid
+===============================
+
+This example demonstrates how to calculate and visualize personalized PageRank on a grid. We use the :meth:`igraph.GraphBase.personalized_pagerank` method, and demonstrate the effects on a grid graph.
```
szhorvat, line 8:
```suggestion
This example demonstrates how to calculate and visualize personalized PageRank on a grid. We use the :meth:`igraph.Graph.personalized_pagerank` method, and demonstrate the effects on a grid graph.
```

Let's link to the `Graph` version of methods, for consistency. `GraphBase` should be treated as an internal class.

doc/examples_sphinx-gallery/personalized_pagerank.py
```diff
@@ -0,0 +1,81 @@
...
+
+    # Plot the graph
+    ig.plot(graph, target=ax, layout=graph.layout_grid())
+
+    # Add a colorbar
+    sm = cm.ScalarMappable(norm=plt.Normalize(min(p_pagerank), max(p_pagerank)), cmap=cmap)
+    plt.colorbar(sm, ax=ax, label="Personalized PageRank")
+
+    plt.title("Graph with Personalized PageRank")
+    plt.show()
```
szhorvat, line 57: Let's make the plot not be stretched vertically or horizontally. Something like `plt.axis("equal")` can achieve this.

doc/examples_sphinx-gallery/personalized_pagerank.py
```diff
@@ -0,0 +1,81 @@
...
+g = ig.Graph.Lattice(dim=grid_size, circular=False)
+
+# %%
+# Then we initialize the ``reset_vector`` (it's length should be equal to the number of vertices in the graph):
+reset_vector = np.zeros(g.vcount())
+
+# %%
+# Then we set the nodes to prioritize, for example nodes with indices ``0`` and ``6``:
+reset_vector[0] = 1
+reset_vector[6] = 1
```
szhorvat, line 73: Let's choose two nodes that are a bit further from each other, and give one of them a higher weight than the other.

Alternatively, use the `reset_vertices` parameter and pass in the vertices. They will have equal weights in this case.

It's your choice which pattern you demonstrate.

doc/examples_sphinx-gallery/personalized_pagerank.py
```diff
@@ -0,0 +1,81 @@
...
+
+# %%
+# Then we calculate the personalized PageRank:
+personalized_page_rank = g.personalized_pagerank(weights=None, damping=0.85, reset=reset_vector)
```
szhorvat, line 77: Let's remove `weights` here since there are no weights to think about. You can choose to tune the damping to make the plot look more interesting. Even if you leave it at the default of 0.85, keep the damping parameter, as this is something people may want to play with.

You could even choose to show two plots with two different damping values, to illustrate the effect of this parameter.

szhorvat (changes requested):
Nice work, well done!

Could you add a very short explanation, perhaps a "note", about what PageRank is? The PageRank score of a vertex is the frequency with which a random walker traverses that vertex. After each step, the walker has a `1 - damping` chance to restart the walk, picking a starting vertices with probabilities given in the `reset` vector.

doc/examples_sphinx-gallery/personalized_pagerank.py
szhorvat: @ntamas In IGraph/M I allow `reset_vertices` to be a dictionary. They keys are vertices and the values are reset weights. What do you think about implementing the same for python-igraph, assuming that @BeaMarton13 is interested in doing it?

BeaMarton13: Absolutely, we can try to do it.

PR #836, opened 2025-06-18
Edge betweenness-based community detection now computes modularity even for weighted graphs

szhorvat:
Because of: 

https://github.com/igraph/igraph/blob/b9efa5de8cf988e432f5b379c0b40a3af62ca395/CHANGELOG.md?plain=1#L36


ntamas:
Thanks!

PR #841, opened 2025-07-26
feat: align_layout()

szhorvat:


szhorvat:
@ntamas, I need help here. How do I pass a graph object to a free function? I'm getting this error from `PyIGraph_ToCGraph`:

```
In [7]: ig.align_layout(g, lay.coords)
TypeError                                 Traceback (most recent call last)
Cell In[7], line 1
----> 1 ig.align_layout(g, lay.coords)

TypeError: expected graph object, got <class 'list'>

In [8]: g
Out[8]: <igraph.Graph at 0x1039b7050>
```

szhorvat:
@ntamas, some questions regarding this PR:

 - What is the proper way to take a `Graph` as the input, in a function implemented in C?  I see at least 3 ways: (1) use `O!` and `igraphmodule_GraphType`, then pick the `igraph_t` pointer out.  (2) Use `PyIGraph_ToCGraph()` — seems the simplest (3) Use `igraphmodule_PyObject_to_igraph_t()` as I do now — seems the most consistent with how things are already done elsewhere
 - Where do we make use of alignment automatically? Options:
     1. Only in `layout_auto()`
     2. In all "organic" layout functions, such as fr, kr, etc. (but of course not in layouts such as tree, bipartite, circle, random, etc.)
     3. In all "organic" layout functions, but also add a parameter `align=True` to them to make it possible to disable alignment
 - Do we align only smallish graphs, e.g. up to a vertex count of 1000? Alignment should be relatively cheap (it's linear time), but it's mostly pointless for messy large graphs, _except_ for making sure that the plot is more wide than tall, which does have some value.

szhorvat:
@ntamas, once again, I have to ask if you can help update the few failing tests ... I expect these must somehow hard-code properties of specific layouts. Unfortunately I wasn't able to run these specific tests locally due to missing dependencies. Running Python tests is extremely painful as it keeps accessing online resourcs, which keep timing out and barely give me 10 kB / s ...

szhorvat:
image

PR #844, opened 2025-08-01
docs: community detection guide

BeaMarton13:
Copilot was used to assist with the generation of some functions and explanatory texts.


szhorvat:
It's starting to look great!

Some tips:

 - Be sure to add necessary packages to the install and update steps of `mkdocs.sh`
 - Can we find a way to add a download button to the end of pages, to allow people to get the notebook?
 - Note that this is integrated into the doc pages, can we link directly to function docs? If this is easy, let's do it, if not, it's not worth spending time on.
 - Check the visual output and use only as many bullet points as looks good. I don't suggest trying to include notebook cells in bullet points. Just write explanations above cells without a bullet.

doc/source/conf.py
```diff
@@ -64,11 +64,16 @@ def get_igraph_version():
     "sphinx.ext.coverage",
     "sphinx.ext.mathjax",
     "sphinx.ext.intersphinx",
+    # NOTE: sphinx_gallery.gen_gallery is not used because it conflicts with pydoctor and nbsphinx
```
ntamas, line 67: Apparently this comment is not relevant any more, right?

doc/source/conf.py
```diff
@@ -64,11 +64,16 @@ def get_igraph_version():
...
     "sphinx_gallery.gen_gallery",
-    #'sphinx_panels',
+    # #'sphinx_panels',
```
ntamas, line 69: Why the double comment?

doc/source/conf.py
```diff
@@ -108,7 +113,7 @@ def get_igraph_version():
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
-exclude_patterns = ["include/*.rst"]
+exclude_patterns = ["include/*.rst", "**.ipynb_checkpoints", "_build", "tutorials/*.ipynb"]
```
ntamas, line 116: Is `_build` really needed in this array? I thought it would be ignored by default.

Also, why do we ignore `tutorials/*.ipynb`?

doc/source/index.rst
```diff
@@ -67,8 +67,9 @@ Documentation
 
        - :doc:`Quick start <tutorials/quickstart>`
        - :doc:`Gallery of examples <tutorials/index>`
+       - :doc:`Community Detection Guide <community_detection_guide>`
```
ntamas, line 70: Let's capitalize the first word only to be consistent with the other headings in this section. (I know it's a matter of style, I'm just suggesting this for sake of consistency)

scripts/mkdoc.sh
```diff
@@ -54,13 +54,52 @@ if [ ! -d ".venv" ]; then
...
-  .venv/bin/pip install -q -U pip wheel sphinx==7.4.7 matplotlib pandas scipy pydoctor sphinx-rtd-theme
+  .venv/bin/pip install -q -U pip wheel sphinx==7.4.7 matplotlib pandas scipy pydoctor sphinx-rtd-theme nbsphinx pandoc sphinx-design ipython ipywidgets
 else
   # Upgrade pip in the virtualenv
   echo "Upgrading pip in virtualenv..."
   .venv/bin/pip install -q -U pip wheel
+
+  # Check if nbsphinx is installed and install it if not
+  echo "Checking for nbsphinx, pandoc, sphinx-design and ipython..."
+  if ! .venv/bin/pip show nbsphinx > /dev/null 2>&1; then
```
ntamas, line 67: This can be simplified with a `for` loop:

```bash
DEPS="nbsphinx pandoc sphinx-design ipython ipywidgets"
for PKG in $DEPS; do
    if ! .venv/bin/pip show ${PKG} > /dev/null 2>&1; then
        echo "$PKG not found. Installing..."
        .venv/bin/pip install -q $PKG
    else
        echo "$PKG is already installed."
    fi
done
```

But maybe it's easier to just let `pip` do the heavy lfiting:

```
pip install -q -U nbsphinx pandoc sphinx-design ipython ipywidgets
```

Don't spend too much time on this; at some point I will rewrite this whole script using `uv` and a proper virtualenv management system anyway.

doc/source/conf.py
BeaMarton13: I just forgot to remove this from an earlier phase, thanks for the heads-up.

ntamas:
@szhorvat What's the status of this, can this be merged into `develop`?

PR #851, opened 2025-08-14
feat: Graph.Nearest_Neighbor_Graph()

szhorvat:
@Zepeacedust, this is an example of exposing a new graph generator in python-igraph, that you can later use as a model to expose the Delaunay generator.

This is not my comfort zone so let's wait for comments from @ntamas.


ntamas (approved):
~~I prefer `Graph.Nearest_Neighbor()` so I'll do the rename.~~

Nope, I changed my mind ;-)

PR #857, opened 2025-09-24
Documentation on plotting with iplotx

iosonofabio:

Hi all,

As you know we support matplotlib as a valid igraph drawing backend because it enables interactive visualisations, animations, editing, etc. It works but not terribly well - bugs, limitations due to design flaws (my own) and back compatibility requirements, etc.

`iplotx` (https://iplotx.readthedocs.org) is de facto a reimplementation of our matplotlib drawing functionality with some improvements:

- Better overall codebase
- Better use of proper matplotlib artists
- Fixed many bugs compared to our internal imple
- More style options, including a style library
- 100% guarantee that igraph and networkx data structures will look identical if they are conceptually identical
- Easier to edit, animate, etc.

(I have talked with many of you about this already, so this PR should be kind of expected too.)

In this PR, I would like to introduce mentions of `iplotx` in our documentation so that users can choose that option if they so wish. For now, I have only added a gallery example (which works locally and *should* work on RTD).

Additional proposals:
- I could add mention of this in our drawing tutorial
- We could deprecate our internal matplotlib backend altogether and make iplotx an optional dependency. This would require some discussion of course. The pros would be we won't have to fix bugs in our codebase anymore, iplotx would take responsibility for that, which leads to better user experience. The cons is that users will need iplotx installed to use igraph with matplotlib. iplotx depends on numpy and pandas (the latter could conceivably be dropped in the future, it is only used internally and  not part of the API).

It would be great if a couple folks could give a little feedback on the general idea of the PR for now. After the first round of feedback, I'd be happy to fix CI bugs, style, etc. etc.

Btw: I just made an identical PR on the NetworkX repo.

Thank you!

### DIFF
--- a/doc/examples_sphinx-gallery/plot_iplotx.py
+++ b/doc/examples_sphinx-gallery/plot_iplotx.py
@@ -1,0 +1,63 @@
+"""
+.. _tutorials-iplotx:
+
+==============================
+Visualising graphs with iplotx
+==============================
+``iplotx`` (https://iplotx.readthedocs.io) is a library for visualisation of graphs/networks
+with direct compatibility with both igraph and NetworkX. It uses ``matplotlib`` behind the
+scenes so the results are compatible with the current igraph matplotlib backend and many
+additional chart types (e.g. bar charts, annotations).
+
+Compared to the standard visualisations shipped with igraph, ``iplotx`` offers:
+
+- More styling options
+- More consistent behaviour across DPI resolutions and backends
+- More consistent matplotlib artists for plot editing and animation
+
+"""
+
+import igraph as ig
+import iplotx as ipx
+
+# Construct a graph with 5 vertices
+n_vertices = 5
+edges = [(0, 1), (0, 2), (0, 3), (0, 4), (1, 2), (1, 3), (1, 4), (3, 4)]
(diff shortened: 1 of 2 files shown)


ntamas:
Looks good to me!

As for deprecating the `matplotlib` backend: this is definitely the preferred way of going forward; however, I would like to release 1.0.0 first from the Python interface, based on version 1.0.0 of the C core, without additional major changes as my resources are limited without funding behind the project. Once version 1.0.0 is released, we need to think about how to move forward. To be honest, I'd rather not invest too much time in the current Python interface and just focus on the new one instead (if we ever find resources for working on it). The new one has no plotting backend at all at the moment so it's pretty much a clean slate and we can just tell people to use `iplotx` for plotting. But all this discussion is moot if we never reach a stage with the new Python interface that it becomes usable.

ntamas:
@iosonofabio Since I want to ship 1.0.0 of the Python interface soon, is it okay if I just merge this PR as-is?

ntamas:
> Would it make sense to add an "iplotx backend" for 1.0?

I don't want to undertake any serious development work before the release of 1.0 so my vote would be a no. We can add a new plotting backend any time after 1.0 if there are resources for it but I don't want the release to be delayed any further.

